Issue #7, opened 2017-05-07
Replace System with FnMut


msiglreith:
The current code for defining a new task looks like this (from the README)
```Rust
struct PrintSystem;

impl<'a> System<'a> for PrintSystem {
    type SystemData = PrintData<'a>;

    fn work(&mut self, bundle: PrintData<'a>) {
        println!("{:?}", &*bundle.a);
        println!("{:?}", &*bundle.b);
        
        *bundle.b = ResB; // We can mutate ResB here
                          // because it's `FetchMut`.
    }
}
```

Every system needs an additional struct (often unit type from my own experience) which becomes quite verbose. Alternative might be replace the Task trait with a function:

```rust
fn system_print(_: &mut (), bundle: PrintData<'a>) {
    println!("{:?}", &*bundle.a);
    println!("{:?}", &*bundle.b);
    *bundle.b = ResB;
}
```

The dispatch builder would then take a `F: FnMut(&mut T, U), U: SystemData` (or something similar).

```rust
let mut dispatcher = DispatcherBuilder::new()
    .add((), system_print, "print", &[]) // Adds a system "print" without dependencies
    .finish();
```

I'm _not_ 100% sure if this will be possible, tests some time ago looked positive..

torkleyy:
@msiglreith I highly support making this more convenient, but this is actually not that easy at the moment because the `FnMut` has to provide an implementation for every possible lifetime `'a`, but you cannot express that without associated types (and `FnMut`'s parameters are type parameters).

torkleyy:
This issue could be solved with Associated Type Constructors (it would be a benefit for the whole crate). @msiglreith If you have any ideas, please don't hesitate to comment!

msiglreith:
@torkleyy I'll give it a try later on!

msiglreith:
Hm, indeed quite challenging!
I got some first results, but need to cleanup and I hope to reduce the complexity a bit ..

torkleyy:
It seems we cannot solve this at the moment. Let's come back to it when we have HKTs. I plan to publish a new version in the next days (fyi, in case anybody wants to make changes).

ebkalderon:
[...]

PR #10, opened 2017-05-10
Replace System trait with functions

msiglreith:
cc #7 

Splitting the current `exec` into two functions: `fetch` which will acquire the data bundle and `exec` which executes the worker function.
The main drawback is the introduction of more `Box` and the requirement of a `unsafe` block, wasn't able to remove this with `Any` due to `'static` restriction. (Helper helps!)

An new constraint is that `Resources` needs to outlive the `Dispatcher` but I think thats pretty minor..

### DIFF
--- a/src/dispatch.rs
+++ b/src/dispatch.rs
@@ -1,10 +1,11 @@
 use std::sync::Arc;
 
 use fnv::FnvHashMap;
-use rayon::{Configuration, Scope, ThreadPool, scope};
+use rayon::{Configuration, ThreadPool, scope};
 
 use bitset::AtomicBitSet;
-use {ResourceId, Resources, System, SystemData};
+use {ResourceId, Resources, SystemData};
+use system::Helper;
 
 #[derive(Default)]
 struct Dependencies {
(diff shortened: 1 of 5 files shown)


src/dispatch.rs
```diff
@@ -300,34 +311,11 @@ impl<'t> DispatcherBuilder<'t> {
...
-
-struct SystemInfo<'t> {
+struct SystemInfo {
     dependents: Vec<usize>,
-    exec: Box<ExecSystem + Send + 't>,
 }
+
+struct SystemWorker<'a> {
+    exec: Box<FnMut(Box<Helper + 'a>, &AtomicBitSet) + Send + Sync>,
+    fetch: Box<Fn(&'a Resources) -> Box<Helper + 'a> + Send + Sync>,
```
msiglreith: Adding the `Scope<'a>` into the closure makes the lifetime story quite difficult (impossible?).
That's why the fetch and exec part got split. The extra box is there to hide the return type.

torkleyy:
@msiglreith Would it be possible to bind `FnMut`'s `SystemData` to a fake lifetime somehow?

msiglreith:
@torkleyy Not really sure what you mean.
The current issue I'm facing is to satisfy the Scope lifetime somehow I think. Executing with spawning new threads works fine when combining the `exec` and `fetch` parts. Then it also looks similar to the current `System` trait based approach.

torkleyy:
Well, `FnMut`s could work a lot like the `System` trait does now, if there was only a way to specify HRTBs for type parameters (which obviously does not make much sense). So if we could just use some fake lifetime, it would be like `where F: FnMut(D), D: SystemData<'fake>` and everything could be implemented just the way systems are. I don't feel very comfortable with this though.

> The current issue I'm facing is to satisfy the Scope lifetime

This was the part of `shred` I spent most of the time implementing it for.

msiglreith:
Rethinking it a bit, and after implementing it for my old `respecs` proof of concept, the approach you mentioned above
> I think we might keep the System trait and have closures additionally. That way, you can use a closure if you don't need per-system data and you can use a struct if you do.

seems to be the way to go here. Fully closure based would be nice to have but seems not be possible without performance loss atm.
Thanks for the dicussion and taking the time!

PR #15, opened 2017-05-20
Fix TrustCell and another critical bug

torkleyy:
Wow, I didn't realize these bugs the first time I wrote this code.

### DIFF
--- a/src/cell.rs
+++ b/src/cell.rs
@@ -1,13 +1,28 @@
 use std::cell::UnsafeCell;
+use std::error::Error;
+use std::fmt::{Display, Error as FormatError, Formatter};
 use std::ops::{Deref, DerefMut};
-#[cfg(debug_assertions)]
 use std::sync::Arc;
-#[cfg(debug_assertions)]
 use std::sync::atomic::{AtomicUsize, Ordering};
+
+#[derive(Clone, Copy, Debug)]
+pub struct InvalidBorrow;
+
+impl Display for InvalidBorrow {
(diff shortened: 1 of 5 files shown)


torkleyy:
@vitvakatu Would you like to review? I'm especially worried if cell is correct now.

src/cell.rs
```diff
@@ -48,72 +61,64 @@ impl<'a, T> DerefMut for RefMut<'a, T> {
...
+            flag: self.flag.clone(),
+            value: unsafe { &mut *self.inner.get() },
+        }
     }
 
-    #[cfg(debug_assertions)]
-    pub unsafe fn borrow_unchecked_mut(&self) -> RefMut<T> {
-        debug_assert_eq!(0, self.flag.load(Ordering::Acquire), "already borrowed");
+    fn check_flag_read(&self) -> Result<(), InvalidBorrow> {
+        loop {
```
vitvakatu: Why do you need `loop` here?

torkleyy: It's a little complicated:

I'd essentially like a `compare_not_equal_and_add`, but that's of course not available. So I first have to fetch the current snapshot. I then check if it's `!0` and if that's the case, there is already a mutable borrow. In case it's not, I compare if it's still that snapshot and if so, I increase the flag by one. This extra check is necessary because there could be a `store` between the first `load` and the `compare_and_swap`. In that case, I have to retry because I do not know the current value anymore.

PR #20, opened 2017-05-30
Remove `Debug` bound for `Resource` trait.

Object905:
When using type from other crate not every type implements `Debug`, and `Debug` is not use everywhere meaningfully anyway.

### DIFF
--- a/src/res.rs
+++ b/src/res.rs
@@ -1,7 +1,6 @@
 //! Module for resource related types
 
 use std::any::TypeId;
-use std::fmt::Debug;
 use std::hash::Hash;
 use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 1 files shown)


torkleyy:
I added the trait bound intentionally. According to the [API guidelines], every type should implement `Debug`. Without forcing the bound, we cannot debug `Resources`.

torkleyy:
cc @kvark @Aceeri

kvark:
Why don't we lift the restriction from `Resource` but still have `derive(Debug)` working only when `Resource: Debug`?

torkleyy:
That's not possible; there is no way to find out if all the types are `Debug` (nor access their vtables) because we have trait objects in `Resources`.

kvark:
To be honest, I find this requirement (to have everything public as `Debug`) to be a little ridiculous. Is there a debugging use case where `Resource: Debug` would really help?

torkleyy:
> To be honest, I find this requirement (to have everything public as Debug) to be a little ridiculous.

It is.

> Is there a debugging use case where Resource: Debug would really help?

Well, just if you want to print all components...which brings me to the idea that we could use @Aceeri's component groups for this. I guess that's a great compromise. Thanks everybody!

PR #21, opened 2017-06-02
Use rayon-core instead of rayon

Object905:
We are not using other rayon parts anyway.


kvark:
Looks good to me! My only concern would be that *if* `ryon_core` is supposed to be somewhat an internal detail of `ryan` then depending on it would be less stable.
@torkleyy ?

PR #22, opened 2017-06-02
Give better control over system dispatching.

Object905:


src/dispatch.rs
```diff
@@ -68,15 +68,16 @@ impl AsyncDispatcher {
...
     /// Waits for all the systems to finish
-    /// and executes thread local systems (if there
-    /// are any).
     pub fn wait(&mut self) {
         self.signal
             .take()
             .expect(ERR_NO_DISPATCH)
             .wait()
             .expect("The worker thread may have panicked");
+    }
```
torkleyy: Why would I want to block for some amount of time but not execute the thread local systems?

src/dispatch.rs
```diff
@@ -68,15 +68,16 @@ impl AsyncDispatcher {
     }
 
     /// Waits for all the systems to finish
```
torkleyy: Should probably have documentation stating that it's not in fact waiting for "all the systems to finish".

src/dispatch.rs
```diff
@@ -169,10 +171,12 @@ impl<'t> Dispatcher<'t> {
 
         #[cfg(target_os = "emscripten")]
         self.dispatch_seq(res);
+
+        self.dispatch_thread_local(res);
```
torkleyy: Can inline that call and remove the method.

src/dispatch.rs
Object905: Because for example i just executed thread local systems just after async system dispatching and do not want to execute them again.

torkleyy: If that's the case, you're *doing it wrong*<sup>*TM*</sup> The dispatcher assumes all the resources are available to the running system, which isn't the case then.

src/dispatch.rs
Object905: I think that this method is actually needed public, because it separates control of thread local systems and others. In my case i don't want to execute thread local systems all the time after executing parallel, but some times I want to execute only them on demand. In that case I could create separate dispatcher, but with this little method it's easier.

torkleyy: I'd just like to avoid having any more contracts here. Could we instead have two methods, `wait()` and `wait_without_tl()` (or a better name) so that valid API usage already makes it correct in this case?

Object905: On it

Object905: Actually it's not the case for `AsyncDispatcher` since you can't execute thread local systems while executing asynchronously, so i might just leave `AsyncDispatcher` as it was.

Object905: It just felt that I deprive AsyncDispatcher from `Dispatcher`s functionality.

torkleyy: > you can't execute thread local systems while executing 

What do you mean by that? Looking at your PR, it seems to me that you *can* execute thread local systems while the other systems are executed asynchronously, which is what I'd like to avoid.

Object905: I meant that in case if i leave AsyncDispatcher as it was.

torkleyy: Okay, so do you like my proposal with two methods? I assumed it would solve your case, because you wrote that you sometimes want to execute the thread-local systems but not always.

Object905: It will do, but my worry is naming. Executing thread local systems in method named `wait` is 
 just implicit.

torkleyy: It's kind of implicit, but I would assume it from the users perspective, because it says it executes all the systems. We can however add a comment explaining that.

src/dispatch.rs
```diff
@@ -29,6 +29,7 @@ pub struct AsyncDispatcher {
 #[cfg(not(target_os = "emscripten"))]
 impl AsyncDispatcher {
     /// Dispatches the systems asynchronously.
+    /// Do not executes thread local systems.
```
torkleyy: nit: Should be "Does not execute [...]"

torkleyy:
It seems you deleted your branch, so I restored it and will open a PR in a second.

PR #23, opened 2017-06-02
Change resource id to usize and make SystemData functions receive resource id

torkleyy:
Using `usize` for ids has several advantages:

* Easier to implement than hashed values because I only need to accept `usize`, no need for two variants (`Hash`-implementor and hashed value)
* No collisions of ids

---

Accepting the additional ids allows to actually have different system data for every id.

### DIFF
--- a/src/dispatch.rs
+++ b/src/dispatch.rs
@@ -388,12 +388,25 @@ impl<'t> DispatcherBuilder<'t> {
     /// # Panics
     ///
     /// * if the specified dependency does not exist
-    pub fn add<T>(mut self, system: T, name: &str, dep: &[&str]) -> Self
+    pub fn add<T>(self, system: T, name: &str, dep: &[&str]) -> Self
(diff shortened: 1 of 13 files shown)


src/dispatch.rs
```diff
@@ -388,12 +388,25 @@ impl<'t> DispatcherBuilder<'t> {
     /// # Panics
     ///
     /// * if the specified dependency does not exist
-    pub fn add<T>(mut self, system: T, name: &str, dep: &[&str]) -> Self
+    pub fn add<T>(self, system: T, name: &str, dep: &[&str]) -> Self
```
Object905: I think that it needs to be documented that the default id is `0`, or make it something like `usize::MAX`

torkleyy: It's documented for `Resources`, `ResourceId` and `Resources::add`, so it should be quite obvious. If you still feel we should add it, I'd be happy to accept a PR.

PR #26, opened 2017-06-03
Add tests to res module

torkleyy:


torkleyy:
Let's try it out then!

bors r+

PR #31, opened 2017-06-03
Add Stages and dispatch tests

torkleyy:
cc @nchashch

This is how I imagined to have those systems divided into stages:

```rust
let dispatcher = DispatcherBuilder::new()
    .add(InputHandler, "handle_input", &[])
    .add(NetworkSynchronizer, "netsync", &[])
    .new_stage()
    .add(ForceGenerator, "force", &[])
    .add(Integrator, "integrator", &["force"])
    .add(EnemySpawner)
    .new_stage()
    .add(Renderer, "render", &[])
    .add(Audio, "audio", &[])
    .build();
```

cc https://github.com/amethyst/amethyst/issues/238

### DIFF
--- a/src/dispatch.rs
+++ b/src/dispatch.rs
@@ -50,6 +50,7 @@ impl AsyncDispatcher {
                                          inner.ready.clone(),
                                          res,
                                          &inner.running,
+                                         &inner.stages,
                                          &mut inner.systems);
                 }
 
@@ -58,13 +59,14 @@ impl AsyncDispatcher {
     }
 
     fn dispatch_inner(d: &Dependencies,
-                      r: Vec<usize>,
+                      r: Vec<Vec<usize>>,
                       resources: Arc<Resources>,
                       run: &AtomicBitSet,
+                      stages: &[usize],
                       sys: &mut Vec<SystemInfo>) {
         let res = &*resources;
 
-        scope(move |s| Dispatcher::dispatch_inner(d, r, res, run, s, sys));
+        Dispatcher::dispatch_inner(d, r, res, run, stages, sys);
     }
 
     /// Waits for all the systems to finish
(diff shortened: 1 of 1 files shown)


torkleyy:
I'm not yet sure about the API design. I tried to keep the builder flat, but I'm not sure whether or not users will like it.

msiglreith:
Would you mind to quickly summarize what a `stage` is and what kind of issues this new API solves?

If I understand it correctly, after each stage there is some artifical 'barrier', so each system of the current stage has to finish before moving to the next stage. If so I honestly don't really see the point of this feature, why we would want limit parallel execution of systems? (I'm not really convinced of the mentioned argument in the issue about easier debugging/profiling)

torkleyy:
> If I understand it correctly, after each stage there is some artifical 'barrier'

Yes, that's correct.

> If so I honestly don't really see the point of this feature, why we would want limit parallel execution of systems?

You'd otherwise specify all the previous systems as dependency which is no better.

> Would you mind to quickly summarize what a stage is and what kind of issues this new API solves?

As you've said, it's a way of sequentializing part of the dispatching. We already have one sort of sequentialization, dependencies. However, as @nchashch explained in the Amethyst issue, some systems should only be executed after many, many other systems executed (for example you only want to render after executing the other systems). This is what this PR is trying to solve.

torkleyy:
@msiglreith Should I rename the new method to `add_barrier` maybe? I think it's a bit more self-explanatory.

vitvakatu:
@torkleyy I'm not sure that `add_barrier` will be better.
What do think about `next_stage`? 
Also I thought a bit about flat builder design. IMO it's okay now.

torkleyy:
@vitvakatu I don't care how we name it, let's wait for somebody else to comment on this PR.

msiglreith:
@torkleyy Thanks for the explanation, helps alot!

> You'd otherwise specify all the previous systems as dependency 

Sounds fair enough. It's indeed probably easier to just setting a barrier at a specific point and move on, without caring too much with all the dependencies for the moment. At later point in the developement users always can go back and optimize their systems.

> Should I rename the new method to add_barrier maybe? I think it's a bit more self-explanatory.

I agree, this is probably easier to understand!

torkleyy:
Thanks everyone! (will cleanup later, got an optimization that is blocked on this)

bors r+

PR #33, opened 2017-06-04
Optimize by introducing stages

### LINKED ISSUE #6: Benchmark and improve dispatching
The current implementation of fetching the next task to execute is very naive (the implementation is about [here](https://github.com/torkleyy/shred/blob/c145e9e86bb0656a7cc66cbaf1f6b092d9eb8e5f/src/dispatch.rs#L80)).

It may be an advantage that, because of the builder, useful information could be pre-computed, although I'm clueless how and what exactly.

torkleyy:
This optimization is big! :tada: Benchmark running time decreased by 74%.

Fixes #6


src/lib.rs
```diff
@@ -58,18 +58,18 @@ extern crate fnv;
 extern crate mopa;
 #[cfg(not(target_os = "emscripten"))]
 extern crate pulse;
+extern crate rayon;
 #[cfg(not(target_os = "emscripten"))]
 extern crate rayon_core;
```
Object905: rayon_core probably not needed anymore?

torkleyy: Given that `rayon_core` is stable and `rayon` is not, we can just use `rayon_core` where possible I guess.

src/dispatch/builder.rs
```diff
@@ -0,0 +1,209 @@
...
+    ///
+    /// # Panics
+    ///
+    /// * if the specified dependency does not exist
+    pub fn add<T>(mut self, system: T, name: &str, dep: &[&str]) -> Self
+        where T: for<'c> System<'c> + Send + 'a
+    {
+        let id = self.id_gen.next();
+        let mut reads: Vec<_> = unsafe { T::SystemData::reads(0) };
+        let writes = unsafe { T::SystemData::writes(0) };
```
Object905: Why are these calls unsafe? Yes if returned ids are invalid it will cause a undefined behavior, but i doesn't mean that it must be marked as unsafe.

torkleyy: You're right, this should be changed. My initial version of shred didn't check borrow rules in release mode, but it now does so..let's remove the unsafe.

torkleyy: It won't be undefined behaviour. If it was, unsafe would be just right.

src/lib.rs
Object905: Well `rayon` itself uses `rayon_core`, and reexports all parts of it, so using this parts from `rayon_core` is all the same.

src/dispatch/builder.rs
Object905: Actually data race is undefined behavior.

torkleyy: Yes, it is, but we don't have any data races, it panics.

src/dispatch/builder.rs
```diff
@@ -0,0 +1,209 @@
...
+    ///
+    /// The default behaviour is to create a thread pool on `finish`.
+    /// If you already have a rayon `ThreadPool`, it's highly recommended to configure
+    /// this builder to use it with `with_pool` instead.
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Adds a new system with a given name and a list of dependencies.
+    /// Please not that the dependency should be added before
```
WaDelma: s/not/note

src/dispatch/builder.rs
```diff
@@ -0,0 +1,209 @@
...
+        where T: for<'c> System<'c> + Send + 'a
+    {
+        let id = self.id_gen.next();
+        let mut reads: Vec<_> = T::SystemData::reads(0);
+        let writes = T::SystemData::writes(0);
+
+        reads.sort();
+        reads.dedup();
+
+        let mut reads = SmallVec::from_vec(reads);
```
WaDelma: Is there a reason that `SystemData::reads` and `SystemData::writes` don't return `SmallVec`s?

src/dispatch/builder.rs
```diff
@@ -0,0 +1,209 @@
...
+    ///
+    /// # Panics
+    ///
+    /// * if the specified dependency does not exist
+    pub fn add<T>(mut self, system: T, name: &str, dep: &[&str]) -> Self
+        where T: for<'c> System<'c> + Send + 'a
+    {
+        let id = self.id_gen.next();
+        let mut reads: Vec<_> = T::SystemData::reads(0);
+        let writes = T::SystemData::writes(0);
```
WaDelma: Shouldn't these `0`s be `id` instead? Or am I missing something?

src/dispatch/builder.rs
```diff
@@ -0,0 +1,209 @@
...
+        let mut reads = SmallVec::from_vec(reads);
+        reads.shrink_to_fit();
+
+        let mut writes = SmallVec::from_vec(writes);
+        writes.shrink_to_fit();
+
+        let dependencies: SystemDependencies = dep.iter()
+            .map(|x| {
+                     *self.map
+                          .get(x.to_owned())
```
WaDelma: Because how `HashMap::get` is defined `&str` can be used for `HashMap<String, _>::get` and `to_owned` is unnessary here.

src/dispatch/builder.rs
torkleyy: Thanks for reviewing, @WaDelma!

My code is correct, however it's not obvious why I did it like that, because I never wrote my design decision down.

First of all, in case you don't know, system data most likely consists of other structs implementing `SystemData`, so for example: 

```rust
#[derive(SystemData)]
struct Data<'a> {
    a: ReadStorage<'a, A>,
    b: WriteStorage<'a, B>,
}
```

Now, I'd like resources to have additional ids (in Specs these were called component ids). So `fetch` / `reads` / `writes` should be parameterized over this resource id. However, the system data directly associated with a system, doesn't need such an id. It just gets passed `0` for that reason.

> Shouldn't these 0s be id instead?

The `id` you're seeing here is the system id, which is totally unrelated to the resource id.

src/dispatch/builder.rs
torkleyy: Two reasons:

* Didn't want to make it a public dependency
* The system data may read from a resource several times. Therefore, the stack storage would often be too small at first, but not necessary anymore later. That's why I do the conversion after removing duplicates

I didn't yet check if it makes sense to use `SmallVec` for reads / writes at all. It's only used for building the stages, after all.

torkleyy:
I got another idea how to optimize this even more, hold on.

torkleyy:
Yeah! I finally got it finished.

src/dispatch/stage.rs
```diff
@@ -0,0 +1,438 @@
...
+              R::IntoIter: Clone,
+              W: IntoIterator<Item = &'rw ResourceId>,
+              W::IntoIter: Clone
+    {
+        let new_reads = new_reads.into_iter();
+        let new_writes = new_writes.into_iter();
+
+        (self.barrier..self.stages.len())
+            .map(|stage| {
+                let conflict = Self::find_conflict(&*self.ids,
```
kvark: any reason this is not just `self.find_conflict(stage, new_reads.clone(), new_writes.clone(), new_dep)`?

src/dispatch/stage.rs
```diff
@@ -0,0 +1,438 @@
...
+                                                   new_reads.clone(),
+                                                   new_writes.clone(),
+                                                   new_dep);
+                self.remove_ids(stage, new_dep);
+                (stage, conflict)
+            })
+            .find(|&(stage, conflict)| match conflict {
+                      Conflict::None => true,
+                      Conflict::Single(group) => {
+                          self.stages[stage].groups[group].len() < MAX_SYSTEMS_PER_GROUP - 1 &&
```
kvark: why `MAX_SYSTEMS_PER_GROUP - 1` ?

src/dispatch/stage.rs
```diff
@@ -0,0 +1,438 @@
...
+                          self.improves_balance(stage, group, new_time as u8)
+                      }
+                      Conflict::Multiple => false,
+                  })
+            .map(|(stage, conflict)| match conflict {
+                     Conflict::None => InsertionTarget::Stage(stage),
+                     Conflict::Single(group) => InsertionTarget::Group(stage, group),
+                     _ => unreachable!(),
```
kvark: let's have `Conflict::Multiple` here explicitly instead of `_`

src/dispatch/stage.rs
```diff
@@ -0,0 +1,438 @@
...
+            .fold(Conflict::None, Conflict::add);
+
+        conflict
+    }
+
+    /// Removes the ids of a given stage from the passed dependency list.
+    fn remove_ids(&self, stage: usize, new_dep: &mut SmallVec<[SystemId; 4]>) {
+        if !new_dep.is_empty() {
+            for id in self.ids[stage].iter().flat_map(|id_group| id_group) {
+                if let Some(index) = new_dep.iter().position(|x| *x == *id) {
```
kvark: is there `SmallVec::retain()`?

src/dispatch/stage.rs
torkleyy: Yes, easier to test.

src/dispatch/stage.rs
torkleyy: The `ArrayVec` only supports up to `MAX_SYSTEMS_PER_GROUP`, so any additional system would just be discarded. So I check if there's enough space left.

src/dispatch/stage.rs
torkleyy: Wanted to use that, but no, unfortunately there isn't.

torkleyy:
bors r+

Thanks @Object905 @WaDelma @kvark for reviewing!

PR #39, opened 2017-06-07
Panic if multiple systems with same name are registered

torkleyy:
Allows `""` to be used in case you don't need a name.

r? @WaDelma


WaDelma:
Yeah looks like that should do it.

torkleyy:
Great! You can use `""` now, so no need 255 allocations any more.

bors r+

PR #40, opened 2017-06-14
Update rayon version

pengowen123:
Updates the rayon dependency version to 0.8.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ arrayvec = "0.3"
 fnv = "1"
 mopa = "0.2"
 pulse = "0.5"
-rayon = { version = "0.7", features = ["unstable"] }
+rayon = { version = "0.8", features = ["unstable"] }
 shred-derive = { path = "shred-derive", version = "0.3" }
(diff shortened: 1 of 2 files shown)


kvark:
Doesn't this require bumping the `shred` version?

pengowen123:
I'm not sure what should happen when a dependency gets upgraded. I think the version will end up being bumped anyway when this change makes it to crates.io, so it should be okay.

torkleyy:
Thanks for the PR @pengowen123!

While `rayon` is an internal dependency of shred, thus doesn't require a version bump, I think we shouldn't rush with upgrading to `0.8` since it would result to a duplicated inclusion of `rayon` in the final binary, because Specs still has `0.7` and updating it there would require a bump.

Please leave it open, so we don't forget to merge it once we're heading to the next release.

pengowen123:
The `rayon` error has been fixed, so I have changed the PR to just include upgrading the `rayon` version. @torkleyy Will do!

kvark:
If it's an internal dependency, I don't think having it duplicated should be much of a concern.

PR #42, opened 2017-06-17
Add `try_*` variants to methods on Resources that return an Option instead of panicking

pengowen123:
Users of the existing methods may wish to handle the error, so the new methods will return `None` in the case of a requested resource not existing.

### DIFF
--- a/src/res.rs
+++ b/src/res.rs
@@ -10,6 +10,8 @@ use mopa::Any;
 use cell::{Ref, RefMut, TrustCell};
 use system::SystemData;
 
+const RESOURCE_NOT_FOUND: &str = "No resource with the given id";
+
 /// Return value of [`Resources::fetch`].
 ///
 /// [`Resources::fetch`]: struct.Resources.html#method.fetch
@@ -249,12 +251,21 @@ impl Resources {
     pub fn fetch<T>(&self, id: usize) -> Fetch<T>
         where T: Resource
     {
-        let c = self.fetch_internal(TypeId::of::<T>(), id);
+        self.try_fetch(id).expect(RESOURCE_NOT_FOUND)
+    }
 
-        Fetch {
-            inner: c.borrow(),
-            phantom: PhantomData,
-        }
+    /// Like `fetch`, but returns an `Option` instead of panicking in the case of the resource
+    /// being accessed mutably.
+    pub fn try_fetch<T>(&self, id: usize) -> Option<Fetch<T>>
+        where T: Resource
+    {
+        self.try_fetch_internal(TypeId::of::<T>(), id)
+            .map(|r| {
+            Fetch {
+                inner: r.borrow(),
+                phantom: PhantomData,
+            }
+        })
     }
 
     /// Fetches the resource with the specified type `T` mutably.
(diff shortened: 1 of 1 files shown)


src/res.rs
```diff
@@ -249,12 +251,22 @@ impl Resources {
     pub fn fetch<T>(&self, id: usize) -> Fetch<T>
         where T: Resource
     {
-        let c = self.fetch_internal(TypeId::of::<T>(), id);
+        self.try_fetch(id).expect(RESOURCE_NOT_FOUND)
```
torkleyy: Shouldn't this use `fetch_internal`?

src/res.rs
```diff
@@ -249,12 +251,22 @@ impl Resources {
...
-        Fetch {
-            inner: c.borrow(),
-            phantom: PhantomData,
-        }
+    /// Like `fetch`, but returns an `Option` instead of panicking in the case of the resource
+    /// being accessed mutably.
+    pub fn try_fetch<T>(&self, id: usize) -> Option<Fetch<T>>
+        where T: Resource
+    {
+        let c = self.try_fetch_internal(TypeId::of::<T>(), id);
```
torkleyy: Let's inline `c` here (and please to the same for the other methods, too).

src/res.rs
```diff
@@ -263,36 +275,71 @@ impl Resources {
     pub fn fetch_mut<T>(&self, id: usize) -> FetchMut<T>
         where T: Resource
     {
-        let c = self.fetch_internal(TypeId::of::<T>(), id);
+        self.try_fetch_mut(id).expect(RESOURCE_NOT_FOUND)
```
torkleyy: Same here, shouldn't it use `fetch_internal`?

src/res.rs
pengowen123: I wanted to avoid duplication of the code that builds the `Fetch`. Do you want me to leave it the way it was?

torkleyy: I see, okay. No need to change it, right.

pengowen123:
I have inlined the variables as requested. Some of the code may not be formatted the way you want, so you may want to run rustfmt.

src/res.rs
```diff
@@ -263,36 +274,67 @@ impl Resources {
...
     }
 
     /// Fetches the resource with the specified type id.
     ///
     /// Please see `fetch` for details.
+    ///
+    /// # Panics
+    ///
+    /// Panics if no resource with the id exists.
     pub fn fetch_id(&self, id: TypeId, comp_id: usize) -> FetchId {
```
torkleyy: Last thing: I think we should stay consistent with `fetch` then, let's use `try_fetch` here and `.expect` the value.

We should be able to get rid of `fetch_internal` then.

pengowen123:
`fetch_internal` has been removed. If there are no other changes to make, should I squash the commits?

src/res.rs
```diff
@@ -263,36 +274,63 @@ impl Resources {
...
     /// Fetches the resource with the specified type id mutably.
     ///
     /// Please see `fetch` for details.
+    ///
+    /// # Panics
+    ///
+    /// Panics if no resource with the id exists.
     pub fn fetch_id_mut(&self, id: TypeId, comp_id: usize) -> FetchIdMut {
-        let c = self.fetch_internal(id, comp_id);
+        let c = self.try_fetch_internal(id, comp_id).expect(RESOURCE_NOT_FOUND);
```
torkleyy: Sorry for requesting changes once more, but I think you forgot to change this one to `try_fetch_id_mut().expect(...)`. And yes, please squash your commits afterwards.

pengowen123:
I did forget that, fixed.

kvark:
Looks like everybody is happy :)
bors r+

kvark:
:(

kvark:
bors - y u no like me?

notriddle:
Apparently, @torkleyy is the only current reviewer for the shred repo.

torkleyy:
I'm sure I changed it last time..

notriddle:
[Here it is, with just you in the list](https://bors-app.herokuapp.com/repositories/128/settings)

torkleyy:
Okay, but now it should work. Thanks.

PR #44, opened 2017-06-25
Include name of system that was not found in panic message

pengowen123:
I have run into this error many times and it's not fun to find the source using `RUST_BACKTRACE`. This should help find the cause of the error quicker and more easily.


torkleyy:
Thank you!

bors r+

PR #46, opened 2017-08-05
A (failed) attempt to implement "Prefetching"

torkleyy:


torkleyy:
The problem is that one can fetch `SystemData<'a>` (with any `'a`) from `Resources`, but not from `Prefetch`, since that is bound by the field of the system struct.

torkleyy:
There is probably a solution do this in current Rust, but it's probably ugly. Or there is a nice one, which I don't see.

examples/prefetch.rs
```diff
@@ -0,0 +1,50 @@
...
+
+        *b = ResB; // We can mutate ResB here
+        // because it's `FetchMut`.
+    }
+}
+
+fn main() {
+    let mut resources = Resources::new();
+    let mut dispatcher = DispatcherBuilder::new()
+        .add(PrintSystem::new(&resources), "print", &[]) // Adds a system "print" without dependencies
```
torkleyy: That's the idea here: we get the cell from the resources on initialization, so there is no need for hash map access anymore.

torkleyy:
Closing because after #69 and #75 the access is pretty fast and this feature can't be implement safely with current Rust.

PR #47, opened 2017-08-26
Allow to check if systems are running, without waiting for them.

Binero:
Useful for when you want to use the available time on the main thread.


torkleyy:
Thanks for your contribution!

bors r+

PR #50, opened 2017-09-09
Fix bug with chained dependencies

torkleyy:
Previously, the depdendency chain

```
A <- B <- C
```

could only be specified by adding an additional
dependency from C to A. Otherwise, it would give you

```
A <-- C
B
```

in the dispatcher.

Fixes slide-rs/specs#253

### DIFF
--- a/src/dispatch/stage.rs
+++ b/src/dispatch/stage.rs
@@ -242,20 +242,41 @@ impl<'a> StagesBuilder<'a> {
         let new_writes = new_writes.into_iter();
 
         let num_groups = ids[stage].len();
+        let mut dep_conflict = false;
 
         let conflict = (0..num_groups)
             .filter(|&group| {
-                let read_and_write = writes[stage][group]
+                let reads_and_writes = writes[stage][group]
                     .iter()
                     .chain(reads[stage][group].iter());
 
-                check_intersection(new_writes.clone(), read_and_write) ||
-                check_intersection(new_reads.clone(), writes[stage][group].iter()) ||
-                check_intersection(new_dep.iter(), ids[stage][group].iter())
+                let inters = check_intersection(new_writes.clone(), reads_and_writes) ||
+                check_intersection(new_reads.clone(), writes[stage][group].iter());
+
+                if inters {
+                    true
+                } else {
+                    if check_intersection(new_dep.iter(), ids[stage][group].iter()) {
+                        dep_conflict = true;
+
+                        true
+                    } else {
+                        false
+                    }
+                }
+
             })
             .fold(Conflict::None, Conflict::add);
 
-        conflict
+        // If there is a dependency in the dependency list
+        // which was not in a previous or this stage,
+        // return `Multiple` conflict.
+
+        if (dep_conflict && new_dep.len() > 1) || (!dep_conflict && !new_dep.is_empty()) {
+            Conflict::Multiple
+        } else {
+            conflict
+        }
     }
 
     /// Removes the ids of a given stage from the passed dependency list.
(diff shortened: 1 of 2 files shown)


torkleyy:
Updated PR to also bump version.

src/dispatch/stage.rs
```diff
@@ -242,20 +242,42 @@ impl<'a> StagesBuilder<'a> {
         let new_writes = new_writes.into_iter();
 
         let num_groups = ids[stage].len();
+        let mut dep_conflict = false;
 
         let conflict = (0..num_groups)
             .filter(|&group| {
                 let read_and_write = writes[stage][group]
```
WaDelma: Wouldn't `read_or_write` be better name for this variable.

src/dispatch/stage.rs
```diff
@@ -242,20 +242,42 @@ impl<'a> StagesBuilder<'a> {
...
             .fold(Conflict::None, Conflict::add);
 
-        conflict
+        // If there is a dependency in the dependency list
+        // which was not in a previous or this stage,
+        // return `Multiple` conflict.
+
+        match conflict {
+            Conflict::None if !new_dep.is_empty() => Conflict::Multiple,
+            Conflict::Single(_) if (dep_conflict && new_dep.len() > 1) || !new_dep.is_empty() => Conflict::Multiple,
```
WaDelma: I am not sure what this is supposed to be doing, but it seems wrong: if `new_dep.len() > 1` is true then `!new_dep.is_empty()` is also true, which means that `(dep_conflict && new_dep.len() > 1)` doesn't do anything.

src/dispatch/stage.rs
```diff
@@ -242,20 +242,42 @@ impl<'a> StagesBuilder<'a> {
...
                     .iter()
                     .chain(reads[stage][group].iter());
 
-                check_intersection(new_writes.clone(), read_and_write) ||
-                check_intersection(new_reads.clone(), writes[stage][group].iter()) ||
-                check_intersection(new_dep.iter(), ids[stage][group].iter())
+                let inters = check_intersection(new_writes.clone(), read_and_write) ||
+                check_intersection(new_reads.clone(), writes[stage][group].iter());
+
+                if !inters {
```
WaDelma: I would avoid using negation when you have if else.

src/dispatch/stage.rs
WaDelma: Also it's kinda annoying that there is `!new_dep.is_empty()` in multiple guards, but I couldn't come up with better formulation.

kvark: Sounds like a lack of expressiveness in `Conflict::Multiple`

torkleyy: Yeah. Maybe I should rename the variant?

src/dispatch/stage.rs
torkleyy: Makes sense, I used negation because the code was lazily evaluated before..

src/dispatch/stage.rs
torkleyy: No, this is an iterator with both read and write ids.

WaDelma: Well I associate and with intersection and or with union, so I first read that the iterator has elements that are both write and read. Maybe I am the only one though.

src/dispatch/stage.rs
torkleyy: @WaDelma Here's how this works:

If no conflict has been found (`Conflict::None`), there is no intersection reads / writes / dependencies. Now, just because there is no intersection, doesn't mean we can insert the system right here.

Here's why: There can still be dependencies to systems, which are e.g. in the next stage. That means, I have to check if the dependency list is empty (note that I remove dependencies once I iterated over them).

The other case where we cannot insert the system in this stage is if there is

i) a single conflict and
ii) that conflict is not the single dependency in the list.

torkleyy: If you want a more general overview, there's a short comment at the beginning of the file.

src/dispatch/stage.rs
torkleyy: Would `reads_and_writes` be clearer for you?

WaDelma: Yeah that would be better.

src/dispatch/stage.rs
WaDelma: Does the latest commit fix it? As there is still the same logic problem with `(dep_conflict && new_dep.len() > 1) || unmet_dependency`.

torkleyy:
@WaDelma Done, requesting approval / further suggestions.

src/dispatch/stage.rs
```diff
@@ -242,20 +242,45 @@ impl<'a> StagesBuilder<'a> {
...
+        // which was not in a previous or this stage,
+        // return `Multiple` conflict.
+
+        // true if there's at least one unmet dependency
+        let unmet_dependency = !new_dep.is_empty();
+
+
+        match conflict {
+            Conflict::None if unmet_dependency => Conflict::Multiple,
+            Conflict::Single(_) if (dep_conflict && new_dep.len() > 1) || unmet_dependency => Conflict::Multiple,
```
WaDelma: This still has the same problem.
The shape of the guard is:
````rust
let a: bool;
let n: usize;
if (a && n > 1) || n > 0 {
    // Multiple conflicts
}
````
and so `a`s value is irrelevant, because `n > 0` is true for all values of `n` for which `n > 1` is true.

torkleyy: Oops, you're right.

torkleyy:
Thanks for the reviews!

bors r+

PR #51, opened 2017-09-10
Add support for thread-local resources

### LINKED ISSUE #45: Thread-local-resources
It would be nice to have thread-local-resources, as in, resources that do not implement Sync. 

It would be useful for systems that need access to things like `gfx::Device`, `gfx::Encoder` or `gfx::pso::buffer::ConstantBuffer`.

torkleyy:
I feel bad for using such a hack..

Somehow fixes #45, although I'm not sure if I can recommend using this.

### DIFF
--- a/examples/thread_local.rs
+++ b/examples/thread_local.rs
@@ -2,7 +2,7 @@ extern crate shred;
 #[macro_use]
 extern crate shred_derive;
 
-use shred::{DispatcherBuilder, Fetch, FetchMut, Resources, System};
+use shred::{DispatcherBuilder, Fetch, FetchMut, Resources, System, ThreadLocal};
 
 #[derive(Debug)]
 struct ResA;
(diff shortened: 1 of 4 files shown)


torkleyy:
I'm not even sure if this is 100% necessary. Allowing this might lead to more systems being thread-local, thus limiting concurrency. My personal preference is letting one (thread-local) system own the thread-local data and handle requests, which allows other systems to still run in parallel.

torkleyy:
This would also require bumping the minimum version to 1.19

Xaeroxe:
Your work here is pretty impressive, but I tend to agree with your assessment.  Resources only available to one thread sounds like an unnecessary complication that will just make the library harder to work with for little gain.

omni-viral:
`ThreadLocal` may confuse people as it usually means "each thread gets a copy".

torkleyy:
Ah maybe. My mental model was "this is a thread-local attribute **and** multiple threads have their own copy". So basically I see thread-local as "not shared between them".

PR #55, opened 2017-09-23
Add Par, Seq, ParSeq, par! and seq!

torkleyy:
This allows dispatching with zero virtual calls and maximum flexibility.

### Example

```rust
let mut dispatcher = ParSeq::new(
    seq![
        par![
            SysA,
            SysWithLifetime(&x),
            seq![
                SysC,
                SysD,
            ],
        ],
        SysB,
        SysLocal(&x as *const u8),
    ],
    &pool,
);
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ rust:
 - beta
 - stable
 - 1.17.0
+env:
+  global:
+    - RUST_MIN_STACK=16777216
 branches:
   only:
     - staging
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,6 @@ rust:
 - beta
 - stable
 - 1.17.0
-env:
-  global:
-    - RUST_MIN_STACK=16777216
 branches:
   only:
     - staging
--- a/examples/par_seq.rs
+++ b/examples/par_seq.rs
@@ -1,0 +1,64 @@
+extern crate rayon;
+#[macro_use(par, seq)]
+extern crate shred;
+
+use rayon::ThreadPool;
+
+use shred::{ParSeq, Resources, System};
+
+macro_rules! impl_sys {
+    ($( $id:ident )*) => {
+        $(
+            impl<'a> ::shred::System<'a> for $id {
+                type SystemData = ();
+                fn run(&mut self, _: Self::SystemData) {
+                    println!(stringify!($id));
+                }
+            }
+        )*
+    };
+}
+
+struct SysA;
+struct SysB;
+struct SysC;
+struct SysD;
+struct SysWithLifetime<'a>(&'a u8);
+struct SysLocal(*const u8);
+
+impl_sys!(SysA SysB SysC SysD SysLocal);
+
+impl<'a, 'b> System<'a> for SysWithLifetime<'b> {
+    type SystemData = ();
+
+    fn run(&mut self, _: Self::SystemData) {
+        println!("SysWithLifetime");
+    }
+}
+
+fn main() {
+    #![cfg_attr(rustfmt, rustfmt_skip)]
+
+    let pool = ThreadPool::new(Default::default()).unwrap();
+
+    let mut res = Resources::new();
+    let x = 5u8;
+
+    let mut dispatcher = ParSeq::new(
+        seq![
+            par![
+                SysA,
+                SysWithLifetime(&x),
+                seq![
+                    SysC,
+                    SysD,
+                ],
+            ],
+            SysB,
+            SysLocal(&x as *const u8),
+        ],
+        &pool,
+    );
+
+    dispatcher.dispatch(&mut res);
+}
--- a/src/dispatch/par_seq.rs
+++ b/src/dispatch/par_seq.rs
@@ -1,10 +1,11 @@
 use std::borrow::Borrow;
 
-use res::Resources;
+use dispatch::util::check_intersection;
+use res::{ResourceId, Resources};
 use system::RunNow;
 use system::System;
 
-use rayon::{ThreadPool, join};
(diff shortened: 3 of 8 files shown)


omni-viral:
This is really great!
Would you have to manually setup systems so they won't try to write same resource simultaneously?

torkleyy:
> Would you have to manually setup systems so they won't try to write same resource simultaneously?

Yes. As the documentation says, if two systems actually do that, there'll be a panic. I might add a function to validate the `ParSeq` dispatcher later.

src/dispatch/par_seq.rs
```diff
@@ -0,0 +1,406 @@
...
+        if pool.current_thread_index().is_none() {
+            pool.join(head, tail);
+        } else {
+            join(head, tail);
+        }
+    }
+
+    fn reads(&self) -> Vec<ResourceId> {
+        let mut reads = self.head.reads();
+        reads.extend(self.tail.reads());
```
kvark: this seems rather inefficient from the allocation point of view, assuming `tail.read()` also allocates a vector that gets thrown out

src/dispatch/par_seq.rs
```diff
@@ -0,0 +1,406 @@
...
+mod tests {
+    use super::*;
+    use std::sync::Arc;
+    use std::sync::atomic::*;
+
+    #[test]
+    fn nested_joins() {
+        let pool = ThreadPool::new(Default::default()).unwrap();
+
+        pool.join(|| join(|| join(|| join(|| (), || ()), || ()), || ()), || ());
```
kvark: heheh

torkleyy:
Since this is a big PR, it would be good to have another review.

r? @WaDelma

src/dispatch/par_seq.rs
```diff
@@ -0,0 +1,406 @@
...
+{
+    /// Creates a new `ParSeq` dispatcher.
+    /// `run` is usually created by using the `par!` / `seq!`
+    /// macros.
+    pub fn new(run: T, pool: P) -> Self {
+        ParSeq { run, pool }
+    }
+
+    /// Dispatches the systems using `res`.
+    /// This involves zero virtual cost.
```
WaDelma: s/involves zero virtual cost/doesn't do any virtual function calls/

src/lib.rs
```diff
@@ -49,6 +49,12 @@
 //! }
 //! ```
 //!
+//! Now, the above dispatcher is pretty easy to setup.
+//! There's also a more flexible and performant one;
+//! once you know about system data and parallelization,
```
WaDelma: I think the ordering of sentences is bit confusing here, but I am not sure how to change it.
````rust
//! Once you are more familiar with how system data and parallelization works,
//! you can take look at more flexible and performant way to dispatch using `ParSeq`.
//! Using it is bit trickier, but it allows dispatching without any virtual function calls.
````
Maybe?

File: examples/par_seq.rs [hunk 5]
```diff
@@ -0,0 +1,64 @@
...
+fn main() {
+    #![cfg_attr(rustfmt, rustfmt_skip)]
+
+    let pool = ThreadPool::new(Default::default()).unwrap();
+
+    let mut res = Resources::new();
+    let x = 5u8;
+
+    let mut dispatcher = ParSeq::new(
+        seq![
```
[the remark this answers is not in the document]
omni-viral: 👍

src/dispatch/par_seq.rs
omni-viral: If the cost is only virtual should we avoid it? 😄

examples/par_seq.rs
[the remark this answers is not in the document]
torkleyy: This is not as obvious as the current syntax. It's much more magic in it. So I prefer how it's done right now.

src/dispatch/par_seq.rs
torkleyy: I feel that you *make* virtual function calls, but I can't tell for sure.

examples/par_seq.rs
```diff
@@ -0,0 +1,64 @@
...
+impl<'a, 'b> System<'a> for SysWithLifetime<'b> {
+    type SystemData = ();
+
+    fn run(&mut self, _: Self::SystemData) {
+        println!("SysWithLifetime");
+    }
+}
+
+fn main() {
+    #![cfg_attr(rustfmt, rustfmt_skip)]
```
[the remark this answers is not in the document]
torkleyy: Not extremely bad, but it made one of the macro invocations a one-liner which I wanted to avoid in this particular example.

.travis.yml
```diff
@@ -5,6 +5,9 @@ rust:
 - beta
 - stable
 - 1.17.0
+env:
+  global:
+    - RUST_MIN_STACK=16777216
```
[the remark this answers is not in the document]
torkleyy: I should be able to remove this line.

torkleyy:
Thanks for the reviews @kvark @omni-viral @WaDelma @WaDelma 

bors r+

src/dispatch/par_seq.rs
```diff
@@ -0,0 +1,406 @@
+use std::borrow::Borrow;
+
+use dispatch::util::check_intersection;
+use res::{ResourceId, Resources};
+use system::RunNow;
+use system::System;
+
+use rayon::{join, ThreadPool};
+
+/// The "leave node" for the `Par` / `Seq` list.
```
WaDelma: leaf node

torkleyy: I'm not sure how I came to write this..

PR #57, opened 2017-09-28
Take immutable references to `Resources`

torkleyy:
and fix soundness issue with `AsyncDispatcher`

### DIFF
--- a/src/dispatch/async.rs
+++ b/src/dispatch/async.rs
@@ -1,3 +1,4 @@
+use std::borrow::Borrow;
 use std::sync::{Arc, Mutex};
 
 use pulse::Signal;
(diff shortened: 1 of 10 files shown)


src/dispatch/async.rs
```diff
@@ -34,7 +35,10 @@ pub fn new_async<'a>(
     }
 }
 
-impl<'a> AsyncDispatcher<'a> {
+impl<'a, R> AsyncDispatcher<'a, R>
```
torkleyy: This type parameter allows to store a struct wrapping `Resources`.

src/dispatch/dispatcher.rs
```diff
@@ -23,9 +23,12 @@ impl<'a, 'b> Dispatcher<'a, 'b> {
     ///
     /// and runs `dispatch_thread_local` afterwards.
     ///
+    /// Please note that this method assumes that no resource
+    /// is currently borrowed. If that's the case, it panics.
```
Xaeroxe: If it still requires exclusive access to the underlying resources I'm not sure what benefit we gain from changing this to be immutable.  I remember our intent here is to permit nested dispatching, but in order for a system to call its own dispatcher it would also require exclusive access to the resources, which it can't guarantee because its own dispatcher is currently accessing them.

torkleyy: You slightly misunderstood that. It doesn't require exclusive access to the resources. It just doesn't allow borrowing resources from `Resources`. You could e.g. (not yet implemented) hold a reference to a `TrustCell` to avoid the `HashMap` cost and still dispatch with those resources. Additionally, the shred API cannot be written so that mutable `Resources` are passed on.

torkleyy:
Thanks for reviewing @Xaeroxe 

bors r+

PR #59, opened 2017-10-11
Standardize builder naming

gaudecker:
This pull request changes the `DispatcherBuilder` API to have non-consuming `add_` methods in addition to the consuming `with_` methods. The old `add_` prefix was replaced by a `with_` prefix.

One open question is whether we want the `add_` methods to return `&mut Self` instead of `()`.

Need for this change arose from amethyst/amethyst#304. When standardizing the builder naming in Amethyst, we need non-consuming methods in order to pass Rusts borrow checker.

This is a breaking change.


torkleyy:
I'm not sure yet when we want to release a new version. Specs 0.10 has been released lately, so I'm gonna wait a bit before doing any version bumps.

bors r+

PR #61, opened 2017-11-30
Fix building on emscripten

Thinkofname:
Noticed the build fail when I was trying specs on emscripten. Added missing cfg's to the parts requiring rayon like that was already done


WaDelma:
Would it be possible to check in CI that emscripten works?

PR #62, opened 2017-12-21
Made rayon reference work on emscripten.

ndarilek:
I don't specifically know why this is needed, but when building under Emscripten via cargo-web, rayon cannot be found without this change. Works fine via regular builds, so I'm not sure what's different in the emscripten build environment.

These changes fix my problem without breaking regular builds. Would be nice to merge and release a new version so I can get specs working in my wasm project. I can't seem to get this working by overriding the dependency in my own Cargo.toml.

Thanks!

### DIFF
--- a/src/dispatch/par_seq.rs
+++ b/src/dispatch/par_seq.rs
@@ -1,3 +1,5 @@
+extern crate rayon;
+
 use std::borrow::Borrow;
 
 use dispatch::util::check_intersection;
@@ -5,7 +7,7 @@ use res::{ResourceId, Resources};
 use system::RunNow;
 use system::System;
 
-use rayon::{join, ThreadPool};
+use self::rayon::{join, ThreadPool};
 
 /// The "leave node" for the `Par` / `Seq` list.
 pub struct Nil;


torkleyy:
Hello @ndarilek, thank you for the PR!

Using `cargo-web` 0.3.1 and a freshly installed `asmjs-unknown-emscripten`, I cannot reproduce any errors. Also, `par_seq` shouldn't even be built under emscripten, since it's using `#[cfg(not(target_os = "emscripten"))]`.

ndarilek:
Ah, hmm, I wonder if the issues are that:

  * I'm using cargo-web 0.4.0. Not sure if that makes a difference--I'm 
new to all of this.
  * I use `cargo web start --target-webasm-emscripten`. Maybe the 
different platform name doesn't enable the right set of features to 
disable par_seq? (I.e. webasm vs. asmjs.)

Thanks.

torkleyy:
Uploaded new version, please try again.

ndarilek:
Seems to work. Thanks for the quick turnaround! :)

PR #69, opened 2018-01-17
Use `fxhash` instead of `fnv`

torkleyy:
`fxhash` is more efficient than `fnv` here.


torkleyy:
r? @omni-viral

bors delegate=omni-viral

torkleyy:
bors r=omni-viral

PR #70, opened 2018-01-19
Add `print_par_seq`

torkleyy:
This allows to quickly get a `ParSeq` structure without double and triple checking dependencies.

### DIFF
--- a/src/dispatch/builder.rs
+++ b/src/dispatch/builder.rs
@@ -1,3 +1,5 @@
+use std::fmt;
+
 use fxhash::FxHashMap;
 
 use dispatch::Dispatcher;
@@ -257,7 +259,7 @@ impl<'a, 'b> DispatcherBuilder<'a, 'b> {
     /// that can be easily used to get the graph using the `seq!` and `par!`
     /// macros. This is only recommended for advanced users.
     pub fn print_par_seq(&self) {
-        self.stages_builder.print_par_seq(&self.map);
+        println!("{:#?}", self);
     }
 
     /// Builds the `Dispatcher`.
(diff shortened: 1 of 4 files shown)


src/dispatch/builder.rs
```diff
@@ -253,6 +253,13 @@ impl<'a, 'b> DispatcherBuilder<'a, 'b> {
         self.thread_pool = Some(pool);
     }
 
+    /// Prints the equivalent system graph
+    /// that can be easily used to get the graph using the `seq!` and `par!`
+    /// macros. This is only recommended for advanced users.
+    pub fn print_par_seq(&self) {
+        self.stages_builder.print_par_seq(&self.map);
+    }
```
WaDelma: Can't this be made into `Debug` implementation? This would allow the user to choose if they want to pretty print it (`{:?}` vs `{:#?}`).

src/dispatch/stage.rs
```diff
@@ -156,6 +157,32 @@ impl<'a> StagesBuilder<'a> {
...
+                println!("\t\tseq![");
+                for system in group {
+                    let system: &SystemId = system;
+
+                    let mut name = map
+                        .iter()
+                        .find(|&(_, id)| *id == *system)
+                        .map(|(name, _)| name)
+                        .unwrap()
+                        .to_string();
```
WaDelma: This works, but I would construct new hash map that has the inverse mapping before the loop.

torkleyy: Yeah I know, but then it's just a one-time print that doesn't have to be super optimized.

src/dispatch/builder.rs
torkleyy: Ha, that's an interesting idea!

src/dispatch/stage.rs
WaDelma: I just think it might be clearer that way.

PR #71, opened 2018-02-04
Disable multithreading for all wasm32 targets.

BartAdv:
With the coming of `wasm32-unknown-unknown` it might be good idea to use `target_arch = "wasm32"` to detect whether to build with multithreading support or not.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,10 +21,18 @@ travis-ci = { repository = "torkleyy/shred" }
 arrayvec = "0.3"
 fxhash = "0.2"
 mopa = "0.2"
-pulse = "0.5"
-rayon = "0.8"
+pulse = { version = "0.5", optional = true }
+rayon = { version = "0.8", optional = true }
 shred-derive = { path = "shred-derive", version = "0.3" }
 smallvec = "0.4"
 
 [dev-dependencies]
 cgmath = "0.15"
+
+[features]
+default = ["parallel"]
+parallel = ["pulse", "rayon"]
+
+[[example]]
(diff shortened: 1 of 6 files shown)


torkleyy:
Thanks for your contribution :)

I would prefer to change this to a `parallel` feature, as you already said in [`slide-rs/specs#341`](https://github.com/slide-rs/specs/issues/341).

BartAdv:
Yeah, I'm all open for suggestions - I just wasn't sure what's the so called "best practice" approach would be. In general, you're right, there might be more reasons to disable multithreading than the target itself, but on the other hand, in this case (wasm32) target itself determine that you just *can't* have multithreading...

BartAdv:
Let's maybe coordinate it here: https://github.com/slide-rs/specs/issues/341#issuecomment-362896108

BartAdv:
I've altered the conditionals to use "parallel" feature.

BartAdv:
Hmm, what's up with the versions though? master is `0.5.0`, when I wanted to adjust `specs` to use the version from this PR,  I noticed failures when running tests, and I noticed the version it picked up was `0.5.2`...

torkleyy:
> Hmm, what's up with the versions though? master is 0.5.0, when I wanted to adjust specs to use the version from this PR, I noticed failures when running tests, and I noticed the version it picked up was 0.5.2, which lives in the v0.5 branch, which is a bit ahead, a bit behind the master...

This branch is essentially the future 0.6 version, just didn't bump the number in `Cargo.toml` yet.

bors r+

PR #74, opened 2018-02-11
Fix async dispatcher

### LINKED ISSUE #72: AsyncDispatcher seems to have synchronization issues
This is the first time this occurred: https://travis-ci.org/slide-rs/shred/jobs/339916332#L606

A test calling `mut_res` panicked :(

cc @kvark I think the problem is that `pulse` is using `Relaxed` ordering, so the `drop` call of the resources `Arc` gets moved after the pulse.

torkleyy:
Fixes #72

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,8 +21,7 @@ travis-ci = { repository = "torkleyy/shred" }
 arrayvec = "0.3"
 fxhash = "0.2"
 mopa = "0.2"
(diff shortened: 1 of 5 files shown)


kvark:
So long, pulse, and thanks for all the sync!

torkleyy:
?

kvark:
This removes pulse dependency for shred, and shred is used specs, for which pulse was made in the first place, so...

PR #75, opened 2018-02-11
Remove resource ids

torkleyy:
Motivation for this change: Fetching with an id other than `0` is only possible outside a system or with a custom struct (which can only fetch a constant id). The plan was to use this for scripting, but resources added by scripts can easily be maintained by another resource, since access of those "script resources" is special anyways. This PR removes the unnecessary complexity and improves `Resources` access.

Resource ids were also a source of confusion in Specs where people thought that resource ids / component ids have something to do with entities.


torkleyy:
bors r=Xaeroxe

PR #76, opened 2018-02-16
Bump rayon version

torkleyy:


torkleyy:
r? @Xaeroxe

bors r=Xaeroxe

I think this doesn't require a version bump, since `ThreadPool` is in core, right?

torkleyy:
I wanted to delegate,

bors delegate=Xaeroxe

Xaeroxe:
This is a breaking change.  A rayon type is exposed in the signature for `DispatcherBuilder::with_pool` so any calls to that method would not compile after releasing this PR.

This otherwise looks good to me so

bors r+

Xaeroxe:
Okay I just reread your comment, this is still a breaking change because core had a version bump too.

torkleyy:
rayon-core is still 1.x though.

Xaeroxe:
Right.  I've gotten so used to version numbers below 1.0 that normal post 1.0 handling didn't register to me.

This technically still breaks semver, because a type used in a signature changed, but honestly I think this is acceptable breakage.

torkleyy:
How does this technically break semver? You cannot assume that you can bump a dependency without having to dowlnload a new minor version of another dependency, so assuming 1.2->1.4 isn't breaking (which is guaranteed by semver) I can do the bump, have an item exposed from that crate and do a non-breaking release. At least that's my understanding.

Xaeroxe:
One could imagine a scenario where they've locked themselves to rayon_core 1.2 with the `=` operator then tried to use types from that locked down version with shred.

torkleyy:
Sure, you can always cause something to break. But that's not semver's guarantee. This way, adding a private field to a struct would be breaking, because the user asserts it's a certain string. With a patch release I just say: You don't have to change any code to use this update. If this needs a bump, then you wouldn't be able to do _any_ version bump of a public dependency at all.

From the semver FAQ (https://semver.org):

> Software that explicitly depends on the same dependencies as your package should have their own dependency specifications and the author will notice any conflicts.

Xaeroxe:
Sounds good to me.  Let's get a patch version out then.

torkleyy:
Okay :)

PR #78, opened 2018-03-25
Add `MetaTable`s

torkleyy:
Still need to add documentation, but the logic is in place already.

### DIFF
--- a/src/cell.rs
+++ b/src/cell.rs
@@ -1,3 +1,7 @@
+//! Helper module for some internals, most users don't need to interact with it.
+
+#![allow(missing_docs)]
+
 use std::cell::UnsafeCell;
 use std::error::Error;
 use std::fmt::{Display, Error as FormatError, Formatter};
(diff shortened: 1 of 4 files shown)


WaDelma:
I am not sure what this is used for, so I can't really do review for it.

torkleyy:
See the docs, it's for registering trait implementations of resources. This will also be used for `AnyStorage`.

torkleyy:
Rebased it. If you don't feel comfortable reviewing this PR, I can just as well wait for others to review it or merge it as-is.

kvark:
What issue is being addressed here?

torkleyy:
There's no specific issue filed, I just talked about it on Gitter a few times.

Here the motivation:

There's sometimes the need to call trait methods of resources where you don't know the specific type. Deleting components using `AnyStorage` is an example for that.

There are three problems with the current architecture:

1) the current solution needs an extra field in `World`, a `Vec` with raw fat pointers of `AnyStorage`s
2) the user cannot access that API to add their own custom component storages
3) for scripting you also want an API like this that allows you to access the trait implementations of arbitrary resources

This PR makes all that consistent allows easy access by the user.

PR #80, opened 2018-04-02
Add panics section to docs

derekdreery:


src/res/mod.rs
```diff
@@ -203,6 +203,11 @@ impl Resources {
     /// Fetches the resource with the specified type `T` mutably.
     ///
     /// Please see `fetch` for details.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the resource doesn't exist.
+    /// Panics if the resource is being accessed.
```
torkleyy: maybe "is already being accessed"

derekdreery:
I've made the suggested change, but I'm going to roll this up with some other documentation suggestions.

PR #81, opened 2018-04-02
Add some docs

derekdreery:
Also make some changes that should have no effect on functionality, but aid readability.

### DIFF
--- a/src/cell.rs
+++ b/src/cell.rs
@@ -1,15 +1,15 @@
 //! Helper module for some internals, most users don't need to interact with it.
 
-#![allow(missing_docs)]
-
 use std::cell::UnsafeCell;
 use std::error::Error;
 use std::fmt::{Display, Error as FormatError, Formatter};
 use std::ops::{Deref, DerefMut};
 use std::sync::atomic::{AtomicUsize, Ordering};
+use std::usize;
 
+/// Marker struct for an invalid borrow error
 #[derive(Clone, Copy, Debug)]
-pub struct InvalidBorrow;
+struct InvalidBorrow;
 
 impl Display for InvalidBorrow {
     fn fmt(&self, f: &mut Formatter) -> Result<(), FormatError> {
@@ -24,6 +24,9 @@ impl Error for InvalidBorrow {
     }
 }
 
+/// An immutable reference to data in a `TrustedCell`.
+///
+/// Access the value via. `std::ops::Deref` (e.g. `*val`)
 #[derive(Debug)]
 pub struct Ref<'a, T: 'a> {
     flag: &'a AtomicUsize,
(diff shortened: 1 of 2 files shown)


src/cell.rs
```diff
@@ -1,15 +1,15 @@
...
 use std::cell::UnsafeCell;
 use std::error::Error;
 use std::fmt::{Display, Error as FormatError, Formatter};
 use std::ops::{Deref, DerefMut};
 use std::sync::atomic::{AtomicUsize, Ordering};
+use std::usize;
 
+/// Marker struct for an invalid borrow error
 #[derive(Clone, Copy, Debug)]
-pub struct InvalidBorrow;
```
derekdreery: This seems to be an implementation detail, since problems always result in panics.

src/cell.rs
```diff
@@ -24,6 +24,9 @@ impl Error for InvalidBorrow {
     }
 }
 
+/// An immutable reference to data in a `TrustedCell`.
+///
+/// Access the value via. `std::ops::Deref` (e.g. `*val`)
```
torkleyy: Please remove the full stop here.

src/cell.rs
```diff
@@ -44,6 +47,9 @@ impl<'a, T> Drop for Ref<'a, T> {
     }
 }
 
+/// A mutable reference to data in a `TrustedCell`.
+///
+/// Access the value via. `std::ops::DerefMut` (e.g. `*val`)
```
torkleyy: Same here

src/cell.rs
```diff
@@ -24,6 +24,9 @@ impl Error for InvalidBorrow {
     }
 }
 
+/// An immutable reference to data in a `TrustedCell`.
```
torkleyy: `TrustCell`

src/cell.rs
```diff
@@ -44,6 +47,9 @@ impl<'a, T> Drop for Ref<'a, T> {
     }
 }
 
+/// A mutable reference to data in a `TrustedCell`.
```
torkleyy: Same here

src/cell.rs
torkleyy: :thinking: Maybe we should add `try_*` variants?

torkleyy: As you mentioned, that would be helpful for FFI

derekdreery: I'll add the try variants :)

derekdreery:
I've made the changes you suggested. Because I needed (Partial)Eq implementations for some tests, I just implemented all standard traits. I think they are the ones that are recommended from the libz blitz :)

src/cell.rs
```diff
@@ -120,9 +207,30 @@ impl<T> TrustCell<T> {
...
+        Ok(RefMut {
+            flag: &self.flag,
+            value: unsafe { &mut *self.inner.get() },
+        })
+    }
+
+    /// Gets read-only access to the inner value, bypassing the Cell.
+    ///
+    /// Absence of write access is checked at compile time.
+    pub fn get(&self) -> &T {
```
torkleyy: That doesn't quite work. There could be a mutable borrow (using `borrow_mut`).

derekdreery:
I've left all the trait impls on the error type, because it makes composing errors easier, if you want to test those for equality.

derekdreery:
:D

PR #82, opened 2018-04-08
Improve performance of `MetaTable`

torkleyy:


torkleyy:
Thanks!

bors r=Rhuagh

PR #88, opened 2018-05-08
Expose RunWithPool, use replaced name value

Aceeri:
Just need this to name the type in places.


Aceeri:
Should be good now.

bors+

PR #89, opened 2018-05-08
Add RunNow implementations for Dispatcher/ParSeq

Aceeri:


Xaeroxe:
I'm kind of confused, what goal are we trying to achieve by doing this?

Aceeri:
Mainly so you can use `T: RunNow<'a>` to take in any dispatcher/system instead of having to specify ParSeq or Dispatcher directly.

Aceeri:
Failing travis should be fixed by #88

PR #90, opened 2018-05-10
Allow runtime `SystemData`

### LINKED ISSUE #79: Allow runtime `SystemData`
`SystemData` currently has associated functions for `read` and `write`. This prohibits creating more dynamic systems that don't specify what they fetch at compile-time.

## Proposal

* add `SystemData::Specifier`
* move `SystemData::read` and `SystemData::write` to `Specifier` and let them take `&self`
* add a `specifier` method to `System` with a default implementation that works for all current cases

## Example

This is what a custom implementation could look like (very primitive).

```rust
struct CustomData {
    // ..
}

impl<'a> SystemData<'a> for CustomData {
    type Specifier = CustomSpec;

    fn fetch(spec: &Self::Specifier, res: &'a Resources) -> Self {
        unimplemented!()
    }
}

struct CustomSpec {
    reads: Vec<ResourceId>,
    writes: Vec<ResourceId>,
}

impl Specifier for CustomSpec {
    fn reads(&self) -> Vec<ResourceId> {
        self.reads.clone()
    }

    fn writes(&self) -> Vec<ResourceId> {
        self.writes.clone()
    }
}
```

## Possible complications

The default implementation for `System::specifier` won't be able to use the usual `Specifier::default` function as you'd expect, because we can't specialize the default for that method for `Specifier: Default`.

## Unresolved questions

* Find a better name for `Specifier`

---

Name suggestions for `Specifier` welcome!

torkleyy:
Fixes #79 

Closes #84 

@tversteeg Sorry, I want to get this finished so I decided to do it myself. Thanks for your help though!

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -77,7 +77,7 @@ const NUM_COMPONENTS: usize = 200;
 
 // --------------
 
-#[derive(SystemData)]
+#[derive(StaticSystemData)]
 struct SpringForceData<'a> {
     pos: Read<'a, PosStorage>,
     spring: Read<'a, SpringStorage>,
@@ -109,7 +109,7 @@ impl<'a> System<'a> for SpringForce {
     }
 }
 
-#[derive(SystemData)]
+#[derive(StaticSystemData)]
 struct IntegrationData<'a> {
     force: Read<'a, ForceStorage>,
     mass: Read<'a, MassStorage>,
@@ -154,7 +154,7 @@ impl<'a> System<'a> for IntegrationSystem {
     }
 }
 
-#[derive(SystemData)]
+#[derive(StaticSystemData)]
 struct ClearForceAccumData<'a> {
     force: Write<'a, ForceStorage>,
 }
--- a/src/system.rs
+++ b/src/system.rs
@@ -93,16 +93,12 @@ pub trait StaticSystemData<'a> {
 }
 
 impl<'a, T> SystemData<'a> for T where T: StaticSystemData<'a> {
-    type Accessor = StaticAccessor<T>;
-}
+    type Accessor = StaticAccessor;
 
-impl<'a, T> Accessor for StaticAccessor<T> where T: StaticSystemData<'a> {
-    fn reads(&self) -> Vec<ResourceId> {
-        T::reads()
-    }
+    fn setup(res: &mut Resources) { }
 
-    fn writes(&self) -> Vec<ResourceId> {
-        T::writes()
+    fn fetch(access: &Self::Accessor, res: &'a Resources) -> Self {
+        T
     }
 }
 
@@ -135,7 +131,7 @@ pub trait SystemData<'a> {
 }
 
 impl<'a, T: ?Sized> SystemData<'a> for PhantomData<T> {
-    type Accessor = StaticAccessor<()>;
+    type Accessor = StaticAccessor;
 
     fn setup(_: &mut Resources) {}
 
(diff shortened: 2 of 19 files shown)


src/system.rs
```diff
@@ -1,7 +1,105 @@
 use std::marker::PhantomData;
+use std::ops::Deref;
 
 use {ResourceId, Resources};
 
+/// A trait for accessing read/write bundles of [`ResourceId`]s in ['SystemData']. This can be used
```
torkleyy: These links don't work

src/system.rs
```diff
@@ -68,19 +165,108 @@ pub trait System<'a> {
...
 }
 
-/// A struct implementing
-/// system data indicates that it
-/// bundles some resources which are
-/// required for the execution.
+/// A static system data that can specify its dependencies at statically (at compile-time).
+/// Most system data is a `StaticSystemData`, the `SystemData` type is only needed for very special
+/// setups.
+pub trait StaticSystemData<'a> {
```
torkleyy: Note if you're reading this PR: this is basically the old `SystemData` trait, you can still use it and the new trait will be implemented for you.

benches/bench.rs
```diff
@@ -77,7 +77,7 @@ const NUM_COMPONENTS: usize = 200;
 
 // --------------
 
-#[derive(SystemData)]
```
WaDelma: It's kinda unfortunate that the common case becomes longer after this, but I don't know how to avoid that.

src/system.rs
```diff
@@ -1,7 +1,105 @@
...
+    fn reads(&self) -> Vec<ResourceId> {
+        Vec::new()
+    }
+    fn writes(&self) -> Vec<ResourceId> {
+        Vec::new()
+    }
+}
+
+/// Either an `Accessor` of the system `T` or a reference to it.
+pub enum AccessorCow<'a, 'b, T>
```
WaDelma: Why does this need it's own `Cow`? Why not use stds?

src/system.rs
```diff
@@ -68,19 +165,108 @@ pub trait System<'a> {
         RunningTime::Average
     }
 
+    /// Return the accessor from the [`SystemData`].
+    fn accessor<'b>(&'b self) -> AccessorCow<'a, 'b, Self> {
+        AccessorCow::Owned(
+            <<Self as System<'a>>::SystemData as SystemData<'a>>::Accessor::try_new()
```
WaDelma: This is bit gnarly, but don't know way to improve it.

benches/bench.rs
torkleyy: Yeah, just mentioned that on Gitter, I don't like it either. I think I'm gonna rename `StaticSystemData` back to `SystemData` and call the other one `DynamicSystemData`

src/system.rs
torkleyy: `std`s requires it to be `Clone`

torkleyy: It's not 100% a copy on write type, because it just means it's either a reference or an owned instance.

src/system.rs
torkleyy: Ah, I think I wanted to use `AccessorTy` for this

src/system.rs
WaDelma: `RefOrOwned`, `OwnedOrNot`, `RefOrNot`, `MaybeOwned`, `Borrowned`.
Cannot think good name for it.

torkleyy:
bors delegate=Rhuagh

PR #93, opened 2018-06-07
Add support for function systems

Xaeroxe:
This allows the creation of systems from mere functions, sidestepping the need for tuple structures and reducing boilerplate in simple system definitions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ rust:
 - nightly
 - beta
 - stable
-- 1.18.0
+- 1.20.0
 branches:
   only:
     - staging
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Please see [the benchmark](benches/bench.rs) for a bigger (and useful) example.
 
 ### Required Rust version
 
-`1.18 stable`
+`1.20 stable`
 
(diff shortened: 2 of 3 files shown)


torkleyy:
As soon as we don't box the closures, I also don't see any need to add an implementation for function pointers.

Xaeroxe:
Credit to @omni-viral for coming up with the clever way of getting around lifetime issues.

Xaeroxe:
I'd love to get rid of the macro too but honestly until @omni-viral came along with that we were convinced this was just impossible.  Do you have an alternative?

torkleyy:
Yeah well it's not impossible, but not trivial. I can implement this in
about a week.

Xaeroxe:
Btw pinging you @torkleyy so you can take a crack at this now that you're back

Xaeroxe:
@torkleyy Most of the time I agree with you but in the Amethyst examples many of our systems are unit structures and could be simplified with this.

Xaeroxe:
Although now that I look at this again I'm not completely sure it's a major syntax improvement.  Let's postpone this until such a time when rustc allows us to implement it without such macros.

PR #95, opened 2018-06-11
Add functions for determining the maximum useful thread count

Xaeroxe:
This could be useful for reducing CPU overhead in Amethyst

### DIFF
--- a/src/dispatch/dispatcher.rs
+++ b/src/dispatch/dispatcher.rs
@@ -94,6 +94,18 @@ impl<'a, 'b> Dispatcher<'a, 'b> {
             sys.run_now(res);
         }
     }
+
+    /// This method returns the largest amount of threads this dispatcher
+    /// can make use of.  Allocating anymore threads than this will use CPU
+    /// without conferring any benefit.  (Unless something is running in
+    /// the threadpool other than the dispatcher.)
+    #[cfg(feature = "parallel")]
+    pub fn max_threads(&self) -> usize {
+        self.stages
+            .iter()
+            .map(|s| s.max_threads())
+            .fold(0, |highest, value| highest.max(value))
+    }
 }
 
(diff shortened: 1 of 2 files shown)


src/dispatch/dispatcher.rs
```diff
@@ -94,6 +94,18 @@ impl<'a, 'b> Dispatcher<'a, 'b> {
...
+    /// This method returns the largest amount of threads this dispatcher
+    /// can make use of.  Allocating anymore threads than this will use CPU
+    /// without conferring any benefit.  (Unless something is running in
+    /// the threadpool other than the dispatcher.)
+    #[cfg(feature = "parallel")]
+    pub fn max_threads(&self) -> usize {
+        self.stages
+            .iter()
+            .map(|s| s.max_threads())
+            .fold(0, |highest, value| highest.max(value))
```
torkleyy: Would rayon permit creating 0 threads?

torkleyy:
I'm not sure I can recommend this as there are other things than the
dispatcher running jobs with the pool. E.g. asset loading

src/dispatch/dispatcher.rs
Xaeroxe: Possibly not, but nonetheless the API report is accurate.  A dispatcher with nothing to do needs no threads.

Xaeroxe:
@torkleyy right, which is why Amethyst would add 1 or 2 to the value reported by this API in order to accomodate that.

Xaeroxe:
@torkleyy quick context for the motivation behind this: https://github.com/amethyst/amethyst/issues/780#issuecomment-396281153

torkleyy:
I saw that, but I'm still not convinced it makes sense to create a pool with less threads than cores, and I don't think it's the solution to the high CPU usage.

PR #103, opened 2018-12-06
Remove `::` prefix from derived implementation.

### LINKED ISSUE #102: Update shred-derive to use in-scope shred name
Instead of hard-coding the `shred-derive` to use `::shred`s, remove the leading `::` double colon so that users can use a re-exported `shred` crate. This means users need to specify `use shred` in modules, but it means they do not need to keep track of which version of `shred` that a higher level library uses, as long as the higher level library re-exports the `shred` crate.

azriel91:
This allows an in-scope re-exported `shred` crate to be used.

Closes #102

---

I'll test this in a larger project, give me a few days to get to it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ members = ["shred-derive"]
 
 [dev-dependencies]
 cgmath = "0.16"
-shred-derive = { path = "shred-derive", version = "0.5.0" }
+shred-derive = { path = "shred-derive", version = "0.6.0" }
 
 [features]
 default = ["parallel"]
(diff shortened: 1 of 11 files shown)


azriel91:
Hm, the changes were more of a "get the user to explicitly `use shred`" instead of a Rust 2018 change. Would you still like it to be behind a feature gate? (can't find a way to detect based on edition)

azriel91:
>that would make it a breaking change

yeap, would mean `shred` 0.8 and `shred-derive` 0.6

> whether shred or the individual symbols should be used

Going by the other crates, I'd go with the individual symbols for consistency. Reminder for myself: even if `shred` was re-exported under a different name, you can still import it `as shred`.

azriel91:
Pushed a change for the latter method

PR #105, opened 2018-12-21
Update HashMaps to use hashbrown HashMap

bdelmas:
A small update of the `FxHashMap` `HashMap` to use the more perfomante `hashbrown` `HashMap`.

[`hashbrown`](https://github.com/Amanieu/hashbrown) is a Rust port of Google's high-performance SwissTable hash map and is around [2 times faster](https://github.com/Amanieu/hashbrown#features) than `FxHashMap`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ cache: cargo
 
 rust:
 - nightly
+- beta
+- stable
 - 1.30.0
 
 branches:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ travis-ci = { repository = "torkleyy/shred" }
 
 [dependencies]
 arrayvec = "0.4"
-fxhash = "0.2"
+hashbrown = "0.1.7"
 mopa = "0.2"
 rayon = { version = "1.0", optional = true }
 smallvec = "0.6"
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ MIT/Apache-2.
 
 ## License
 
-`shred` is distributed under the terms of both the MIT 
+`shred` is distributed under the terms of both the MIT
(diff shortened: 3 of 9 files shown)


Xaeroxe:
Benchmark shows this is an improvement, of about 6% for the library, that's good enough for me, though I'd like to know what our new minimum rust version is after this upgrade.

Xaeroxe:
Ok, so the error for 1.20 is because hashbrown used the `..=` operator, which was stabilized in 1.26.  That's likely to be our new minimum version.

Xaeroxe:
Ok, and I think in order to compile for this error we'll need 1.27

LucioFranco:
why even track a min version, just track stable 🤷‍♂️

Xaeroxe:
Not every one has the luxury of `rustup update stable`.  It's useful info for them.

LucioFranco:
@Xaeroxe I'm curious has this actually come up and what situations could someone not `rustup update`?

Xaeroxe:
According to the [Rust 2018 survey results](https://blog.rust-lang.org/2018/11/27/Rust-survey-2018.html) about 10% of users don't install via rustup, with most of that crowd being in "Linux distribution package".  For example, the Ubuntu rustc package is at 1.28.

Xaeroxe:
1.30 is our new minimum.

torkleyy:
Hmm, I'm not super happy with bumping to 1.30 already. Crates should at least stay 2 below stable. I'll merge this with 1.32 out, though. Thanks for the PR!

bdelmas:
Awesome! Thank to you guys! 🎄

bdelmas:
@torkleyy Yes sure I do that as soon as I can!

PR #109, opened 2019-01-17
Renamed 'Resource' to 'World'

TimonPost:
I renamed  'Resource' to 'World'  as stated in [this issue](https://github.com/slide-rs/specs/issues/523).

This PR is part 1 of two PR's described in the linked issue.

### DIFF
--- a/src/res/mod.rs
+++ b/src/res/mod.rs
@@ -109,13 +109,13 @@ impl ResourceId {
 ///
 /// # Resource Ids
 ///
-/// Resources are identified by `ResourceId`s, which consist of a `TypeId`.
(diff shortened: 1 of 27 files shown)


src/res/mod.rs
```diff
@@ -109,13 +109,13 @@ impl ResourceId {
 ///
 /// # Resource Ids
 ///
-/// Resources are identified by `ResourceId`s, which consist of a `TypeId`.
+/// World are identified by `ResourceId`s, which consist of a `TypeId`.
```
torkleyy: This one doesn't make sense

TimonPost: good catch!

PR #121, opened 2019-03-29
Add `{Dispatcher, RunNow, System}::dispose`

### LINKED ISSUE #119: Add `System::dispose`
Requested by @Frizi.

`System::dispose` would be called by `Dispatcher::dispose` and can be used to clean up externally allocated resources.

## Unresolved

* Signature: `fn dispose(&self, world: &World)` / `fn dispose(&self, world: &mut World)` / `fn dispose(&mut self, world: &World)` / `fn dispose(&mut self, world: &mut World)`
* Is it mandatory to call it? How to make sure it gets called?

torkleyy:
Fixes #119

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -125,7 +125,7 @@ pub trait RunNow<'a> {
     /// Performs clean up that requires resources from the `World`.
     /// This commonly removes components from `world` which depend on external
     /// resources.
-    fn dispose(&mut self, world: &mut World);
+    fn dispose(&mut self, world: &mut World) {}
 }
 
 impl<'a, T> RunNow<'a> for T
--- a/src/system.rs
+++ b/src/system.rs
@@ -121,6 +121,11 @@ pub trait RunNow<'a> {
 
     /// Sets up `World` for a later call to `run_now`.
     fn setup(&mut self, world: &mut World);
+
+    /// Performs clean up that requires resources from the `World`.
+    /// This commonly removes components from `world` which depend on external
+    /// resources.
+    fn dispose(&mut self, world: &mut World);
 }
 
 impl<'a, T> RunNow<'a> for T
@@ -134,6 +139,10 @@ where
 
     fn setup(&mut self, world: &mut World) {
         T::setup(self, world);
+    }
+
+    fn dispose(&mut self, world: &mut World) {
+        T::dispose(self, world);
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/system.rs
```diff
@@ -121,6 +121,11 @@ pub trait RunNow<'a> {
 
     /// Sets up `World` for a later call to `run_now`.
     fn setup(&mut self, world: &mut World);
+
+    /// Performs clean up that requires resources from the `World`.
+    /// This commonly removes components from `world` which depend on external
+    /// resources.
+    fn dispose(&mut self, world: &mut World);
```
LucioFranco: prob want a default impl here and an impl in the blanket? This should prob also reduce the amount of breaking thatll happen. I assume its safe to just do nothing here as a default?

src/system.rs
```diff
@@ -185,6 +194,13 @@ pub trait System<'a> {
     fn setup(&mut self, world: &mut World) {
         <Self::SystemData as DynamicSystemData>::setup(&self.accessor(), world)
     }
+
+    /// Performs clean up that requires resources from the `World`.
+    /// This commonly removes components from `world` which depend on external
+    /// resources.
+    fn dispose(&mut self, world: &mut World) {
+        let _ = world;
```
LucioFranco: do you actually need to do this?

torkleyy: If I want no warning and keep `world` (instead of `_world`) as a parameter, yes :)

LucioFranco: why not just `_world`?

src/system.rs
torkleyy: I'm not sure. The `RunNow` API is pretty much for consuming only, not for implementing.

src/system.rs
torkleyy: It shows up in the docs ^^

LucioFranco: but blanket impls you dont care about the fn's generally you just care about that the bounds for T work for you no? I mean this is minor i dont mind either way 😅

torkleyy: This is not a blanket impl, it's the `System` trait. I just wanted to have a nice parameter show up in the docs :sweat_smile:

LucioFranco: Oh my god! You're right I missed that haha, I guess I didnt expand the diff hahaha. you're right this works! Sorry!

torkleyy:
bors delegate=Frizi

Frizi:
Haven't had a chance to review it before, but it does look very good :) Just a question. Why is `dispose` setting a flag instead of consuming the dispatcher? I think it's easier to work with that way, but it's also not entirely safe.

torkleyy:
Thanks for taking a look! That's a good idea, yeah. Totally missed that.

PR #124, opened 2019-03-31
Make dispose self consuming

torkleyy:
cc @Frizi

### DIFF
--- a/src/dispatch/dispatcher.rs
+++ b/src/dispatch/dispatcher.rs
@@ -28,9 +28,6 @@ impl<'a, 'b> Dispatcher<'a, 'b> {
     /// external resources. It is common this method removes components and
     /// / or resources from the `World` which are associated with external
     /// resources.
-    ///
-    /// Calling any method after `dispose` (including `dispose` itself) will
-    /// panic.
     pub fn dispose(self, world: &mut World) {
         for stage in self.stages {
             stage.dispose(world);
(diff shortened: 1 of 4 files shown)


src/dispatch/dispatcher.rs
```diff
@@ -151,8 +131,8 @@ impl<'a, 'b, 'c> RunNow<'a> for Dispatcher<'b, 'c> {
         self.setup(world);
     }
 
-    fn dispose(&mut self, world: &mut World) {
-        self.dispose(world);
+    fn dispose(self: Box<Self>, world: &mut World) {
```
LucioFranco: will this bump min rust version?

torkleyy: No, it won't, that's stable since 1.2x and we're now on 1.32.

torkleyy: Also, that's ensured by CI.

src/dispatch/dispatcher.rs
```diff
@@ -34,14 +31,12 @@ impl<'a, 'b> Dispatcher<'a, 'b> {
     ///
     /// Calling any method after `dispose` (including `dispose` itself) will
```
Frizi: This comment is no longer necessary. You can't call any method anyway because the object was moved :)

src/dispatch/dispatcher.rs
LucioFranco: huh I thought self-referential types was like 1.3X 🤷‍♂️

torkleyy: That was just for things **other than `Box`**

torkleyy:
Thank you!

bors r=LucioFranco,Frizi

PR #127, opened 2019-04-03
Implement World::remove

### LINKED ISSUE #107: Add 'remove' method to Resources.
It seems to be desirable to take ownership again of a resource, without needing to wrap that resource in an Option, by removing it from Resources. For example, in my case, gfx::Device requires to be passed ownership of some things to destroy them, such as buffers, which I am using inside of a Resource. It is possible to wrap these in Option and take() them, but because during dispatch these should never be taken and never be None, it seems, though a very minor inconvenience, an unfortunate solution.

torkleyy:
Fixes #107

### DIFF
--- a/src/cell.rs
+++ b/src/cell.rs
@@ -94,6 +94,11 @@ impl<T> TrustCell<T> {
         }
     }
 
+    /// Consumes this cell and returns ownership of `T`.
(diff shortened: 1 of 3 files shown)


torkleyy:
Thanks!
bors r=distransient,Frizi

distransient:
image

PR #128, opened 2019-04-03
Implement dynamic resource IDs + example

### LINKED ISSUE #94: Cannot create dynamic `ResourceId`s
The [documentation](https://docs.rs/shred/0.7.0/shred/struct.ResourceId.html) states that a `ResourceId` is:
>The id of a Resource, which is a tuple struct with a type id and an additional resource id (represented with a usize).

However, it only contains a `TypeId`, which means there's no way to dynamically construct a `ResourceId`!

torkleyy:
Fixes #94

### DIFF
--- a/examples/dynamic_resource_id.rs
+++ b/examples/dynamic_resource_id.rs
@@ -152,11 +152,9 @@ impl<'a> DynamicSystemData<'a> for ScriptingResData<'a> {
                 .iter()
                 .map(|id| {
                     world
-                        .try_fetch_internal(id.clone())
+                        .try_fetch_by_id(id.clone())
                         .expect("Resource no longer exists")
-                        .borrow()
                 })
-                .map(|r| unsafe { Fetch::from_inner_unchecked(r) })
                 .collect(),
         }
     }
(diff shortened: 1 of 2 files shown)


torkleyy:
Still gotta add some more tests.

torkleyy:
bors try

Moxinilian:
I’d recommend using the word Dynamic instead of Scripting, as otherwise I think it would be too precise to serve it well.

Have you assessed the costs for the current workflow (performance and API-wise)?

torkleyy:
> I’d recommend using the word Dynamic instead of Scripting, as otherwise I think it would be too precise to serve it well.

Oh, you mean in the example? I've used "dynamic" everywhere else AFAIK.

> Have you assessed the costs for the current workflow (performance and API-wise)?

I've done some benchmarking, there was no noticeable performance cost.

API-wise - it's definitely not a high-level API, but it's a very flexible low level API. It should be easy to define your scripting framework with that.

torkleyy:
Thank you!
bors r=Moxinilian,jojolepro

PR #134, opened 2019-04-20
Make Fetch cloneable

Marwes:
Working on allowing [gluon](https://github.com/gluon-lang/gluon) to work with specs/shred and ended up needing to clone some `Fetch<EntitiesRes>` values. I don't know fully how specs/shred work internally but this should be equivalent to doing multiple `fetch` operations and should therefore be sound.

### DIFF
--- a/src/cell.rs
+++ b/src/cell.rs
@@ -233,6 +233,16 @@ mod tests {
     }
 
     #[test]
+    fn allow_clone_reads() {
+        let cell: TrustCell<_> = TrustCell::new(5);
+
+        let a = cell.borrow();
+        let b = a.clone();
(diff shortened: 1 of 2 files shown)


Marwes:
Repo is at https://github.com/Marwes/gluon_specs, example usage is still to be written though I had something working before starting to refactor it.

Marwes:
Added a test and relaxed the ordering to `Release` since we don't care about the fetched value.

I think `Relaxed` _might_ work as well as `Drop` uses `Release`. But I am not entirely clear if a `Relaxed` `add` is guaranteed to be done before the `Release` `sub`

PR #136, opened 2019-05-14
Update derive to use Resources instead of World

shelbyd:
I needed to do this to compile locally. Not sure what the correct approach is in general.


torkleyy:
@shelbyd Hey! I'm sorry, this is currently a little confusing. I changed `Resources` to `World`, and released new `shred` and `shred-derive` versions, but not new Specs version yet. So you're likely using the latest version of `shred-derive`, which is not correct for Specs 0.14.

PR #138, opened 2019-05-23
Remove box indirection in `Fetch{,Mut}`

barskern:
I know that `cell` is a small part of this library that is not really meant for consumption. However upon closer inspection, in my process to learn amethyst, I figured that I might have a change that could cause a tiny performance improvement that hopefully would ripple through-out all consumers of this library. 

## Situation

The existing implementation of `Ref{,Mut}` had the implicit bound `T: Sized`. However internally the `T` was stored behind a reference, which means that it doesn't need to be sized to make `Ref{,Mut}` a sized struct. By removing this bound we can make `T` a `dyn Resource` instead of a `Box<Resource>`, which means that we remove a layer of indirection. This means that instead of having a `Fetch{,Mut}` with a `Ref{,Mut}<'_, Box<Resource>>`, we can rather give it `Ref{,Mut}<'_, dyn Resource>`.

## Solution

To accomplish this I defined `map` on `Ref{,Mut}` (as is implemented in std for `Ref{,Mut}` of `RefCell`) so that one could modify the contents of `Ref{,Mut}`. That way one can work around the fact that the `T` in `TrustCell` has to be sized, by rather changing the reference after it's creation. 

## Performance

I ran the benchmarks in the repo to see a small performance _increase_ in the `basic` case. 

##### Box-indirection (before)

```
test basic          ... bench:       9,152 ns/iter (+/- 90)
test bench_fetching ... bench:       4,602 ns/iter (+/- 7)
```

##### Direct reference (after)

```
test basic                  ... bench:       8,510 ns/iter (+/- 60)
test bench_fetching         ... bench:       4,608 ns/iter (+/- 95)
```

----

Further I added two benchmarks which also highlight the difference.

##### Comparison when accessing value of Ref multiple times

```
test bench_direct_refs      ... bench:       5,010 ns/iter (+/- 50)
test bench_indirection_refs ... bench:       6,618 ns/iter (+/- 182)
```

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -10,7 +10,7 @@ use std::ops::{Index, IndexMut};
 
 use cgmath::Vector3;
 use shred::*;
-use test::Bencher;
+use test::{black_box, Bencher};
 
 #[derive(Debug)]
 struct VecStorage<T> {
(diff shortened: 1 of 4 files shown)


torkleyy:
Looks good, `meta.rs` only uses `try_fetch_internal`, which remains unchanged. :shipit: 
bors r+

barskern:
Thank you very much for the kind feedback, @torkley. I already had a great experience contributing to amethyst a while ago with your help. Hence it's very nice to feel like I'm able to give something back. 😊😊

PR #144, opened 2019-06-21
Batch processing

AndreaCatania:
# Batch dispatching

The **Batch** is a feature that will allow to control the dispatching of a group of systems.
It can run these systems multiple times.

So is possible to register some systems under the Batch, and for this reason they are called Sub Systems. 
Despite this name they are just **Systems** nothing more, nothing less.

The Batch will control it's dispatching, it's setup, and it's dependencies.

To control the Batch execution exist a **BatchController**. It is a trait that a System must implement if he want to control the execution of the Batch.
Then is possible to set it using one of the two: `BatchBuilder::with_batch_controller` or `BatchBuilder::set_batch_controller`.

The **BatchBuilder** is responsible for the creation of the **Batch**.
It works similarly to the **DispatcherBuilder**, and indeed is possible to set the systems and the barriers.
But despite the similarities, it at the end will create a Batch.

Note that depending on the dependencies of the Sub Systems, the Batch can run in parallel with other Systems.
In addition the Sub Systems can run in parallel within the Batch.

So everything in line with the actual shred!

----
Here the test project that I've used to test the code 
[shred_test.zip](https://github.com/slide-rs/shred/files/3360288/shred_test.zip)

### DIFF
--- a/src/dispatch/batch_builder.rs
+++ b/src/dispatch/batch_builder.rs
@@ -1,0 +1,496 @@
+use hashbrown::HashMap;
+
+use crate::{
+    dispatch::{
+        dispatcher::SystemId,
+        stage::{Stage, StagesBuilder},
+    },
+    system::RunningTime,
+    world::ResourceId,
+    RunNow, System, SystemData, World,
+};
+
+/// The BatchController is the additional trait that a normal System must implement
+/// in order to be used as BatchController.
+///
+/// # Example
+///
+/// ```
+/// use shred::{BatchController, System};
+///
+/// /// The following controller restart the execution of the Batch three times.
+/// #[derive(Default)]
+/// pub struct TestBatchControllerSystem{
+///     iterations: i32,
+/// }
+///
+/// impl BatchController for TestBatchControllerSystem{
+///
+///     fn prepare_dispatching(&mut self){
+///         self.iterations = 0;
+///     }
+///     
+///     fn want_to_dispatch(&mut self) -> bool{
+///         self.iterations < 3
+///     }
+/// }
+///
+/// impl<'a> System<'a> for TestBatchControllerSystem {
+///
+///     type SystemData = ();
+///
+///     fn run(&mut self, (): Self::SystemData){
+///
+///         self.iterations += 1;
+///     }
+///
+/// }
+///
+/// ```
+///
+/// Use the function `BatchBuilder::with_batch_controller` or `BatchBuilder::set_batch_controller`
+/// to set your BatchControllerSystem.
+pub trait BatchController {
+    /// This function is called at the start of the BatchSystem execution.
+    fn prepare_dispatching(&mut self);
+
+    /// This control the internal dispatching of the sub systems.
+    ///
+    /// The following pseudo code show how it works.
+    ///
+    /// >loop{
+    /// >    if batch_controller.want_to_dispatch() == false {
+    /// >        break
+    /// >    }
+    /// >
+    /// >    substep Dispatching
+    /// >}
+    ///
+    fn want_to_dispatch(&mut self) -> bool;
+}
+
+pub trait BatchControllerSystem<'a>: BatchController + RunNow<'a> {}
+
+impl<'a, T> BatchControllerSystem<'a> for T where T: BatchController + RunNow<'a> {}
+
+/// This executes only 1 time the BatchSystem.
+/// Use this controller doesn't make much
+/// sense for this reason an error message will be print.
+#[derive(Default)]
+pub struct DefaultBatchControllerSystem {
+    is_first_execution: bool,
+}
+
+impl BatchController for DefaultBatchControllerSystem {
+    fn prepare_dispatching(&mut self) {
+        self.is_first_execution = true;
+    }
+
+    fn want_to_dispatch(&mut self) -> bool {
+        self.is_first_execution
+    }
+}
+
+impl<'a> System<'a> for DefaultBatchControllerSystem {
+    type SystemData = ();
+
+    fn run(&mut self, (): Self::SystemData) {
+        self.is_first_execution = false;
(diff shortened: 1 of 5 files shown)


src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// Please note that this method assumes that no resource
+    /// is currently borrowed. If that's the case, it panics.
+    pub fn dispatch_seq(&mut self, world: &World) {
+        for stage in &mut self.stages {
+            stage.execute_seq(world);
+        }
+    }
+}
+
+impl<'a> RunNow<'a> for Batch<'a>{
```
torkleyy: `RunNow` is not sufficient if you want to add `Batch` to a dispatcher.

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,296 @@
...
+    },
+    world::ResourceId,
+    system::RunningTime,
+    System,
+    RunNow,
+    World,
+};
+
+#[derive(Default)]
+pub struct BatchBuilder<'a>{
```
torkleyy: I'm a bit worried we now have so much code duplicated.

src/dispatch/batch_builder.rs
AndreaCatania: I'm going to work on this soon today.

The solution to this problem is not use the batch system and register the systems under the BatchBuilder directly inside the stages vector.

The problem with this is that using directly the BatchSystem is much more convenient since their systems can't be mixed with the other systems. Also its restart happen locally and no further changes are required.

Instead the solution of placing the batching systems directly inside the stages vector make necessary keep track of them in order to not mix these with the other systems.
This is necessary because a normal system may be executed many times unexpectedly.

What's necessary to change in order to allow to add an object which implement just the RunNow  trait?

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,496 @@
...
+    /// is currently borrowed. If that's the case, it panics.
+    pub fn dispatch_seq(&mut self, world: &World) {
+        for stage in &mut self.stages {
+            stage.execute_seq(world);
+        }
+    }
+}
+
+impl<'a> System<'a> for BatchSystem<'_> {
+    type SystemData = UncheckedWorld<'a>;
```
torkleyy: This doesn't quite work, because now the `BatchSystem` has no resource requirements, thus can run in parallel with everything else (which it shouldn't).

You need to collect all the `reads` and `writes` of all inner system, sort them and remove all the duplicates. Then, you can return those resource ids using a

* custom [`Accessor`](https://docs.rs/shred/0.9.2/shred/trait.Accessor.html) implementation (a struct with two `Vec`s, which you just return from `Accessor::{reads, writes}`)
* custom [`DynamicSystemData`](https://docs.rs/shred/0.9.2/shred/trait.DynamicSystemData.html), which just defines the associated `Accessor` and sets it to your custom implementation

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,496 @@
...
+/// In addition the Sub Systems can run in parallel within the Batch.
+pub struct BatchBuilder<'a> {
+    current_id: usize,
+    map: HashMap<String, SystemId>,
+    stages_builder: StagesBuilder<'a>,
+    read_resources: Vec<ResourceId>,
+    write_resources: Vec<ResourceId>,
+    running_time: f32,
+    system_count: u32,
+    batch_controller: Box<for<'c> BatchControllerSystem<'c> + Send + 'a>,
```
torkleyy: I don't quite see why this should be controlled like this (it also seems like a special case, which shouldn't be in shred).

Instead, execution should be controlled by wrapping the `BatchSystem` (or, preferably, the `Dispatcher`, see my other comment).

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,496 @@
...
+/// It has the capability to restart its systems many times, and this
+/// can be controlled by the user.
+///
+/// The BatchSystem keep track of its internal dependecies and is able to run
+/// in parallel with other systems.
+/// At the same time the subsystems can run in parallel within the BatchSystem.
+///
+/// To add a new batch system you have to call the function `with_batch` which
+/// accept a `BatchBuilder` object that will take care to construct the `BatchSystem`.
+pub struct BatchSystem<'a> {
```
torkleyy: Do we even need this? This looks exactly like `Dispatcher`, with the exception of `batch_controller`, which I don't think should be in here.

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,496 @@
...
+    /// Please note that this method assumes that no resource
+    /// is currently borrowed. If that's the case, it panics.
+    pub fn dispatch_seq(&mut self, world: &World) {
+        for stage in &mut self.stages {
+            stage.execute_seq(world);
+        }
+    }
+}
+
+impl<'a> System<'a> for BatchSystem<'_> {
```
torkleyy: This would then become a `impl<'a> System<'a> for Dispatcher<'_>`.

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,496 @@
...
+            #[cfg(feature = "parallel")]
+            self.dispatch_par(world.0);
+
+            #[cfg(not(feature = "parallel"))]
+            self.dispatch_seq(world.0);
+        }
+    }
+
+    fn running_time(&self) -> RunningTime {
+        self.running_time
```
torkleyy: This should just always return `RunningTime::VeryLong`.

AndreaCatania:
@torkleyy Thanks a lot for the feedback.
I did a new PR since it was easier for me rework from the scratch.
https://github.com/slide-rs/shred/pull/147

PR #146, opened 2019-06-28
Print type name for cell borrow errors

### LINKED ISSUE #140: Print resource type in case of a double borrow

torkleyy:
Fixes #140


torkleyy:
Thanks for taking a look!
bors r+

PR #147, opened 2019-07-05
Implemented batch dispatching

AndreaCatania:
This is the second iteration to implement the Batch dispatching following the feedback received in this PR: https://github.com/slide-rs/shred/pull/144 (Was easier for me redo everything rather change it)

This is the project to test this feature: 
[shred_test.zip](https://github.com/slide-rs/shred/files/3363344/shred_test.zip)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,3 +73,4 @@ name = "thread_local"
 [[example]]
 name = "batch_dispatching"
 required-features = ["parallel"]
+
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,3 +69,7 @@ name = "seq_dispatch"
 
 [[example]]
 name = "thread_local"
+
+[[example]]
+name = "batch_dispatching"
+required-features = ["parallel"]
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -1,6 +1,6 @@
 //! This example shows how to use and define a batch dispatcher.
 //!
-//! The Batch is a feature that allow to control the dispatching of a group of
+//! The batch feature allows to control the dispatching of a group of
 //! systems.
 //!
 //! Specifically here we have three Systems
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -1,4 +1,4 @@
-//! This example show how to use the Batch dispatcher.
+//! This example shows how to use and define a batch dispatcher.
 //!
 //! The Batch is a feature that allow to control the dispatching of a group of
 //! systems.
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -105,7 +105,7 @@ pub struct CustomBatchControllerSystem<'a, 'b> {
 
 impl<'a, 'b> BatchController<'a, 'b> for CustomBatchControllerSystem<'a, 'b> {
     // Using this `BatchBuilderData = ()` make the dispatcher to panic since the run
-    // function will fetch the TomatoStore like the SayHelloSystem does.
+    // function will fetch the `TomatoStore` like the `SayHelloSystem` does.
     // type BatchSystemData = ();
     type BatchSystemData = (Read<'a, TomatoStore>);
 
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -20,13 +20,11 @@
 //! Check the **CustomBatchControllerSystem** which execute its inner Systems
 //! three times.
 
-
 use shred::{
     AccessorCow, BatchAccessor, BatchController, BatchUncheckedWorld, Dispatcher,
     DispatcherBuilder, Read, RunningTime, System, World, Write,
 };
-use std::thread::sleep;
-use std::time::Duration;
+use std::{thread::sleep, time::Duration};
 
 fn main() {
     let mut dispatcher = DispatcherBuilder::new()
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -91,7 +91,7 @@ impl<'a> System<'a> for BuyPotatoSystem {
 pub struct BuyTomatoSystem;
 
 impl<'a> System<'a> for BuyTomatoSystem {
-    type SystemData = (Write<'a, PotatoStore>);
+    type SystemData = (Write<'a, TomatoStore>);
 
     fn run(&mut self, _data: Self::SystemData) {
         println!("Buy Tomato")
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -1,0 +1,150 @@
+//! This example show how to use the Batch dispatcher.
+//!
+//! The Batch is a feature that allow to control the dispatching of a group of systems.
+//!
+//! Specifically here we have three Systems
+//! - **SayHelloSystem**: Which is directly registered under the main dispatcher.
+//! - **BuyTomatoSystem** and **BuyPotatoSystem** are registered in the batch.
+//!
+//! Notice that none of them is using any direct dependency.
+//! The SayHelloSystem is requesting the resources **TomatoStore** and **PotatoStore**,
+//! which are also requested by the other two systems inside the batch and by the batch
+//! controller itself.
+//!
+//! This to demonstrate that the dispatcher is able to correctly organize, their execution.
+//!
+//! Is also showed how to use a custom BatchControllerSystem.
+//! Check the **CustomBatchControllerSystem** which execute its inner Systems three times.
+//!
+
+use shred::{
+    AccessorCow, BatchAccessor, BatchBuilder, BatchController, BatchUncheckedWorld, Dispatcher,
+    DispatcherBuilder, Read, RunningTime, System, World, Write,
+};
+use std::thread::sleep;
+use std::time::Duration;
+
+fn main() {
+    let mut dispatcher = DispatcherBuilder::new()
+        .with(SayHelloSystem, "say_hello_system", &[])
+        .with(
+            BatchBuilder::new(
+                DispatcherBuilder::new()
+                    .with(BuyTomatoSystem, "buy_tomato_system", &[])
+                    .with(BuyPotatoSystem, "buy_potato_system", &[]),
+            )
+            .build::<CustomBatchControllerSystem>(),
+            "BatchSystemTest",
+            &[],
+        )
+        .build();
+
+    let mut world = World::empty();
+
+    dispatcher.setup(&mut world);
+
+    // Running phase
+    for i in 0..10 {
+        println!("Dispatching {} ", i);
+
+        dispatcher.dispatch(&world);
+        sleep(Duration::new(0, 100000000));
+    }
+
+    // Done
+    println!("Execution finished");
+}
+
+// Resources
+
+#[derive(Default)]
+pub struct PotatoStore(i32);
+
+#[derive(Default)]
+pub struct TomatoStore(f32);
+
+// Say Hello Sytem
+
+pub struct SayHelloSystem;
+
+impl<'a> System<'a> for SayHelloSystem {
+    type SystemData = (Write<'a, PotatoStore>, Write<'a, TomatoStore>);
+
+    fn run(&mut self, _data: Self::SystemData) {
+        println!("Hello!")
+    }
+}
+
+// BuyPotatoSystem
+
+pub struct BuyPotatoSystem;
+
+impl<'a> System<'a> for BuyPotatoSystem {
+    type SystemData = (Write<'a, PotatoStore>);
+
+    fn run(&mut self, _data: Self::SystemData) {
+        println!("Buy Potato")
+    }
+}
+
+// Buy Tomato System
+
+pub struct BuyTomatoSystem;
+
+impl<'a> System<'a> for BuyTomatoSystem {
+    type SystemData = (Write<'a, PotatoStore>);
+
+    fn run(&mut self, _data: Self::SystemData) {
+        println!("Buy Tomato")
+    }
+}
+
+// TEST custom Batch Controller
+
+pub struct CustomBatchControllerSystem<'a, 'b> {
+    accessor: BatchAccessor,
+    dispatcher: Dispatcher<'a, 'b>,
+}
+
+impl<'a, 'b> BatchController<'a, 'b> for CustomBatchControllerSystem<'a, 'b> {
+    // Using this `BatchBuilderData = ()` make the dispatcher to panic since the run
+    // function will fetch the TomatoStore like the SayHelloSystem does.
+    // type BatchSystemData = ();
+    type BatchSystemData = (Read<'a, TomatoStore>);
+
+    fn create(accessor: BatchAccessor, dispatcher: Dispatcher<'a, 'b>) -> Self {
+        CustomBatchControllerSystem {
+            accessor,
+            dispatcher,
+        }
+    }
+}
+
+impl<'a> System<'a> for CustomBatchControllerSystem<'_, '_> {
+    type SystemData = BatchUncheckedWorld<'a>;
+
+    fn run(&mut self, data: Self::SystemData) {
+        {
+            // Unload before dispatch the inner dispatcher
+            let _ts = data.0.fetch::<TomatoStore>();
+        }
+        println!("Batch execution");
+        for _i in 0..3 {
+            self.dispatcher.dispatch(data.0);
+        }
+    }
+
+    fn running_time(&self) -> RunningTime {
+        RunningTime::VeryLong
+    }
+
+    fn accessor<'c>(&'c self) -> AccessorCow<'a, 'c, Self> {
+        AccessorCow::Ref(&self.accessor)
+    }
+
+    fn setup(&mut self, world: &mut World) {
+        self.dispatcher.setup(world);
+    }
+}
+
+unsafe impl<'a, 'b> Send for CustomBatchControllerSystem<'a, 'b> {}
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -9,7 +9,7 @@
 //! - `BuyTomatoSystem` and `BuyPotatoSystem` are registered to the batch.
 //!
 //! Notice that none of these systems are directly depending on others.
-//! The SayHelloSystem is requesting the resources **TomatoStore** and
+//! The `SayHelloSystem` is requesting the resources `TomatoStore` and
 //! **PotatoStore**, which are also requested by the other two systems inside
 //! the batch and by the batch controller itself.
 //!
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -4,7 +4,7 @@
 //! systems.
 //!
 //! Specifically here we have three Systems
-//! - **SayHelloSystem**: Which is directly registered under the main
+//! - `SayHelloSystem`: Which is directly registered under the main
 //!   dispatcher.
 //! - `BuyTomatoSystem` and `BuyPotatoSystem` are registered to the batch.
 //!
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -14,7 +14,6 @@
 //! the batch and by the batch controller itself.
 //!
 //! This to demonstrate that the dispatcher is able to correctly organize, their
-//! execution.
 //!
 //! Is also showed how to use a custom BatchControllerSystem.
 //! Check the **CustomBatchControllerSystem** which execute its inner Systems
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -72,7 +72,7 @@ impl<'a> System<'a> for SayHelloSystem {
     }
 }
 
-// BuyPotatoSystem
+/// System that says "Buy Potato"
 
 pub struct BuyPotatoSystem;
 
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -20,23 +20,21 @@
 //! Check the **CustomBatchControllerSystem** which execute its inner Systems
 //! three times.
 
-use std::{thread::sleep, time::Duration};
 
 use shred::{
-    AccessorCow, BatchAccessor, BatchBuilder, BatchController, BatchUncheckedWorld, Dispatcher,
+    AccessorCow, BatchAccessor, BatchController, BatchUncheckedWorld, Dispatcher,
     DispatcherBuilder, Read, RunningTime, System, World, Write,
 };
+use std::thread::sleep;
+use std::time::Duration;
 
 fn main() {
     let mut dispatcher = DispatcherBuilder::new()
         .with(SayHelloSystem, "say_hello_system", &[])
-        .with(
-            BatchBuilder::new(
-                DispatcherBuilder::new()
-                    .with(BuyTomatoSystem, "buy_tomato_system", &[])
-                    .with(BuyPotatoSystem, "buy_potato_system", &[]),
-            )
-            .build::<CustomBatchControllerSystem>(),
+        .with_batch::<CustomBatchControllerSystem>(
+            DispatcherBuilder::new()
+                .with(BuyTomatoSystem, "buy_tomato_system", &[])
+                .with(BuyPotatoSystem, "buy_potato_system", &[]),
             "BatchSystemTest",
             &[],
         )
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -16,7 +16,7 @@
 //! This example demonstrates that the batch dispatcher is able to affect on how the systems inside the batch are executed
 //!
 //! Is also showed how to use a custom BatchControllerSystem.
-//! Check the **CustomBatchControllerSystem** which execute its inner Systems
+//! This is done by defining `CustomBatchControllerSystem` which executes its inner `System`s
 //! three times.
 
 use shred::{
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -104,7 +104,7 @@ pub struct CustomBatchControllerSystem<'a, 'b> {
 }
 
 impl<'a, 'b> BatchController<'a, 'b> for CustomBatchControllerSystem<'a, 'b> {
-    // Using this `BatchBuilderData = ()` make the dispatcher to panic since the run
+    // Leaving `BatchBuilderData` to `()` would make the dispatcher to panic since the run
     // function will fetch the `TomatoStore` like the `SayHelloSystem` does.
     // type BatchSystemData = ();
     type BatchSystemData = (Read<'a, TomatoStore>);
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -15,7 +15,6 @@
 //!
 //! This example demonstrates that the batch dispatcher is able to affect on how the systems inside the batch are executed
 //!
-//! Is also showed how to use a custom BatchControllerSystem.
 //! This is done by defining `CustomBatchControllerSystem` which executes its inner `System`s
 //! three times.
 
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -6,7 +6,7 @@
 //! Specifically here we have three Systems
 //! - **SayHelloSystem**: Which is directly registered under the main
 //!   dispatcher.
-//! - **BuyTomatoSystem** and **BuyPotatoSystem** are registered in the batch.
+//! - `BuyTomatoSystem` and `BuyPotatoSystem` are registered to the batch.
 //!
 //! Notice that none of these systems are directly depending on others.
 //! The SayHelloSystem is requesting the resources **TomatoStore** and
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -1,28 +1,31 @@
 //! This example show how to use the Batch dispatcher.
 //!
-//! The Batch is a feature that allow to control the dispatching of a group of systems.
+//! The Batch is a feature that allow to control the dispatching of a group of
+//! systems.
 //!
 //! Specifically here we have three Systems
-//! - **SayHelloSystem**: Which is directly registered under the main dispatcher.
+//! - **SayHelloSystem**: Which is directly registered under the main
+//!   dispatcher.
 //! - **BuyTomatoSystem** and **BuyPotatoSystem** are registered in the batch.
 //!
 //! Notice that none of them is using any direct dependency.
-//! The SayHelloSystem is requesting the resources **TomatoStore** and **PotatoStore**,
-//! which are also requested by the other two systems inside the batch and by the batch
-//! controller itself.
+//! The SayHelloSystem is requesting the resources **TomatoStore** and
+//! **PotatoStore**, which are also requested by the other two systems inside
+//! the batch and by the batch controller itself.
 //!
-//! This to demonstrate that the dispatcher is able to correctly organize, their execution.
+//! This to demonstrate that the dispatcher is able to correctly organize, their
+//! execution.
 //!
 //! Is also showed how to use a custom BatchControllerSystem.
-//! Check the **CustomBatchControllerSystem** which execute its inner Systems three times.
-//!
+//! Check the **CustomBatchControllerSystem** which execute its inner Systems
+//! three times.
+
+use std::{thread::sleep, time::Duration};
 
 use shred::{
     AccessorCow, BatchAccessor, BatchBuilder, BatchController, BatchUncheckedWorld, Dispatcher,
     DispatcherBuilder, Read, RunningTime, System, World, Write,
 };
-use std::thread::sleep;
-use std::time::Duration;
 
 fn main() {
     let mut dispatcher = DispatcherBuilder::new()
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -10,7 +10,7 @@
 //!
 //! Notice that none of these systems are directly depending on others.
 //! The `SayHelloSystem` is requesting the resources `TomatoStore` and
-//! **PotatoStore**, which are also requested by the other two systems inside
+//! `PotatoStore`, which are also requested by the other two systems inside
 //! the batch and by the batch controller itself.
 //!
 //! This to demonstrate that the dispatcher is able to correctly organize, their
--- a/examples/batch_dispatching.rs
+++ b/examples/batch_dispatching.rs
@@ -8,7 +8,7 @@
 //!   dispatcher.
 //! - **BuyTomatoSystem** and **BuyPotatoSystem** are registered in the batch.
 //!
-//! Notice that none of them is using any direct dependency.
+//! Notice that none of these systems are directly depending on others.
 //! The SayHelloSystem is requesting the resources **TomatoStore** and
 //! **PotatoStore**, which are also requested by the other two systems inside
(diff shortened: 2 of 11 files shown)


WaDelma:
Could the project be an example? As in add it to the examples folder.

AndreaCatania:
Done!

examples/batch_dispatching.rs
```diff
@@ -0,0 +1,150 @@
...
+//!
+//! Is also showed how to use a custom BatchControllerSystem.
+//! Check the **CustomBatchControllerSystem** which execute its inner Systems three times.
+//!
+
+use shred::{
+    AccessorCow, BatchAccessor, BatchBuilder, BatchController, BatchUncheckedWorld, Dispatcher,
+    DispatcherBuilder, Read, RunningTime, System, World, Write,
+};
+use std::thread::sleep;
```
torkleyy: Those should go above `shred` imports (the order for imports typically is 1. `std` 2. `some_crate` 3. `crate::` 4. `self::`).

Also, please run `cargo +nightly fmt` so those get nicely grouped together.

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
+use crate::{
+    dispatch::{Dispatcher, DispatcherBuilder},
+    world::ResourceId,
+    Accessor, AccessorCow, DynamicSystemData, RunningTime, System, SystemData, World,
+};
+
+/// The `BatchBuilder` is responsible for the creation of the Batch.
+///
+/// The `Batch` is a Systam that have the capabilities to contain another dispatcher
```
torkleyy:
```suggestion
/// The `Batch` is a `System` which contains a `Dispatcher`.
```

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+/// Thanks to this characteristic is possible to customize its dispatching
```
torkleyy:
```suggestion
/// By wrapping a `Dispatcher` inside a system, we can control the execution of a whole
/// group of system, without sacrificing parallelism or conciseness.
```

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// The BatchBuilder accept the dispatcher builder as parameter, and the type of the
```
torkleyy:
```suggestion
/// The `BatchBuilder` accepts the `DispatcherBuilder` as parameter, and the type of the
```

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+/// System that will drive the execution of the internal dispatcher.
```
torkleyy:
```suggestion
/// `System` that will drive the execution of the internal dispatcher.
```

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// Note that depending on the dependencies of the SubSystems the Batch
```
torkleyy: "SubSystems" implies some kind of type; also, "dependencies" should probably be clarified by using "resource dependencies" instead.

On dependencies, we should also document cross-dispatcher dependencies don't work.

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+/// can run in parallel with other Systems.
+/// In addition the Sub Systems can run in parallel within the Batch.
+#[derive(Default)]
```
torkleyy: Should also derive `Debug`.

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+        let accessor = BatchAccessor::new(reads, writes);
+        let dispatcher = self.dispatcher_builder.build();
+
+        T::create(accessor, dispatcher)
+    }
+}
+
+/// The BatchAccessor is responsible for the notification of the read and write resources
+/// of the subsystems
+pub struct BatchAccessor {
```
torkleyy: Should implement `Debug`.

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+        writes.dedup();
+        
+        let accessor = BatchAccessor::new(reads, writes);
+        let dispatcher = self.dispatcher_builder.build();
+
+        T::create(accessor, dispatcher)
+    }
+}
+
+/// The BatchAccessor is responsible for the notification of the read and write resources
```
torkleyy:
```suggestion
/// The `BatchAccessor` is used to notify the main dispatcher of the read and write resources
```

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+/// of the subsystems
```
torkleyy:
```suggestion
/// of the `System`s contained by the `Batch` ("sub systems").
```

src/dispatch/mod.rs
```diff
@@ -2,10 +2,11 @@
 pub use self::async_dispatcher::AsyncDispatcher;
 #[cfg(feature = "parallel")]
 pub use self::par_seq::{Par, ParSeq, RunWithPool, Seq};
-pub use self::{builder::DispatcherBuilder, dispatcher::Dispatcher};
+pub use self::{batch_builder::*, builder::DispatcherBuilder, dispatcher::Dispatcher};
```
torkleyy: Please don't use glob import; we don't want all of the types to be exposed (for example, `UncheckedWorld` probably shouldn't be visible).

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    fn writes(&self) -> Vec<ResourceId> {
+        self.writes.clone()
+    }
+}
+
+/// The BatchUncheckedWorld wrap an instance of the world.
+/// It's safe to be used only in the context of the Batch.
+pub struct BatchUncheckedWorld<'a>(pub &'a World);
```
torkleyy: Why is this field `pub`?

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+    type Accessor = BatchAccessor;
+
+    fn setup(_accessor: &Self::Accessor, _world: &mut World) {}
+
+    fn fetch(_access: &Self::Accessor, world: &'a World) -> Self {
+        BatchUncheckedWorld(world)
+    }
+}
+
+/// The BatchController is the additional trait that a normal System must implement
```
torkleyy: Please use backticks for all types.

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+/// in order to be used as BatchControllerSystem.
+pub trait BatchController<'a, 'b> {
+    
+    /// Here you must set all the Resources type that you want to use inside the
```
torkleyy:
```suggestion
    /// Here you must set all the resource types that you want to use inside the
```

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// BatchControllertSystem.
```
torkleyy:
```suggestion
    /// `BatchControllerSystem`.
```

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// All other are fetched automatically.
```
torkleyy:
```suggestion
    /// All the system data needed for the controlled systems are fetched automatically.
```

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+    type BatchSystemData: SystemData<'a>;
+    
+    /// Create the instance of the BatchControllerSystem
+    fn create(accessor: BatchAccessor, dispatcher: Dispatcher<'a, 'b>) -> Self;
```
torkleyy: Any reason why this is called `create` instead of `new`?

src/dispatch/batch_builder.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// Note that depending on the dependencies of the SubSystems the Batch
+/// can run in parallel with other Systems.
+/// In addition the Sub Systems can run in parallel within the Batch.
+#[derive(Default)]
+pub struct BatchBuilder<'a, 'b> {
+    dispatcher_builder: DispatcherBuilder<'a, 'b>,
+}
+
+impl<'a, 'b> BatchBuilder<'a, 'b> {
```
torkleyy: Hmm, on a second thought, we could probably just add a method `build_batch` to the `DispatcherBuilder`, then we don't even need this extra type.

AndreaCatania: Ok I'll add it

src/dispatch/batch_builder.rs
AndreaCatania: Usually `new` is implemented by a struct to initialize itself so I thought was better differentiate it to make it more clear.
Do you think that new is better?

examples/batch_dispatching.rs
AndreaCatania: using cargo nightly fmt put them in this order:
```
use shred::{
    AccessorCow, BatchAccessor, BatchBuilder, BatchController, BatchUncheckedWorld, Dispatcher,
    DispatcherBuilder, Read, RunningTime, System, World, Write,
};
use std::{thread::sleep, time::Duration};
```

is it ok?

src/dispatch/batch_builder.rs
AndreaCatania: This is pub because when you define the custom BatchControllerSystem you want to set it as SystemData

AndreaCatania:
@torkleyy I did all the changes that you advised.

### ThreadPool
About the ThreadPool assignment problem that I mentioned you before, I found a solution that seems nice and elegant.

I've created a struct, called `ThreadPoolWrapper`, which is shared across all batches.
This wrapper contains the actual ThreadPool that now can be modified at any point.

The result of this addition is that when in amethyst the Application struct create the thread pool and assign it to the main DispatcherBuilder it change the internal part of the wrapper and everyone is automatically updated.

To me this seems clear enough but do you think that exist a better way to handle it?

### Tests
I written three tests that demonstrate what you suggested, please let me know if I have to add more.

examples/batch_dispatching.rs
```diff
@@ -0,0 +1,149 @@
...
+        println!("Buy Potato")
+    }
+}
+
+// Buy Tomato System
+
+pub struct BuyTomatoSystem;
+
+impl<'a> System<'a> for BuyTomatoSystem {
+    type SystemData = (Write<'a, PotatoStore>);
```
WaDelma: Is this supposed to be `PotatoStore`?

src/dispatch/batch.rs
```diff
@@ -0,0 +1,418 @@
...
+}
+
+/// The `BatchController` is the additional trait that a normal System must
+/// implement in order to be used as `BatchControllerSystem`.
+pub trait BatchController<'a, 'b> {
+    /// Here you must set all the resource types that you want to use inside the
+    /// `BatchControllerSystem`.
+    /// All the system data needed for the controlled systems is fetched
+    /// automatically.
+    type BatchSystemData: SystemData<'a>;
```
WaDelma: Is there a reason this trait isn't `BatchController<'a, 'b>: System<'a>` and just use `<Self as System<'a>>::SystemData`?

src/dispatch/batch.rs
```diff
@@ -0,0 +1,418 @@
...
+    fn accessor<'c>(&'c self) -> AccessorCow<'a, 'c, Self> {
+        AccessorCow::Ref(&self.accessor)
+    }
+
+    fn setup(&mut self, world: &mut World) {
+        self.dispatcher.setup(world);
+    }
+}
+
+unsafe impl<'a, 'b> Send for DefaultBatchControllerSystem<'a, 'b> {}
```
WaDelma: What makes this un-Send by default? And why is it safe to implement it? I would add comment here.

src/dispatch/batch.rs
```diff
@@ -0,0 +1,418 @@
+use crate::{
+    dispatch::Dispatcher, world::ResourceId, Accessor, AccessorCow, DynamicSystemData, RunningTime,
+    System, SystemData, World,
+};
+
+/// The `BatchAccessor` is used to notify the main dispatcher of the read and
+/// write resources of the `System`s contained by the `Batch` ("sub systems").
+#[derive(Debug)]
+pub struct BatchAccessor {
```
WaDelma: This seems more generic that just batch related: It just holds `reads` and `writes` that cannot be changed inside it. Not sure if it's worth renaming.

src/dispatch/dispatcher.rs
```diff
@@ -2,13 +2,18 @@ use smallvec::SmallVec;
 
 use crate::{dispatch::stage::Stage, system::RunNow, world::World};
 
+/// This wrapper is used to share a replaceable ThreadPool with other
+/// dispatchers. Useful with batch dispatchers.
+#[derive(Debug, Default)]
+pub struct ThreadPoolWrapper(pub Option<::std::sync::Arc<::rayon::ThreadPool>>);
```
WaDelma: Does this need to be newtype? Because the field is public it could be just type alias.

src/dispatch/dispatcher.rs
```diff
@@ -74,11 +79,17 @@ impl<'a, 'b> Dispatcher<'a, 'b> {
...
-            for stage in stages {
-                stage.execute(world);
-            }
-        });
+        self.thread_pool
+            .read()
+            .unwrap()
+            .0
+            .as_ref()
+            .unwrap()
```
WaDelma: There are  multiple places that have this kind of boilerplate. Could this be made helper function?

src/dispatch/stage.rs
```diff
@@ -139,6 +139,34 @@ pub struct StagesBuilder<'a> {
...
 impl<'a> StagesBuilder<'a> {
+    pub fn fetch_all_reads(&self) -> Vec<ResourceId> {
+        let mut v = Vec::<ResourceId>::new();
+        for stages in self.reads.iter() {
+            for groups in stages.iter() {
+                for r in groups.iter() {
+                    v.push(r.clone());
+                }
+            }
+        }
```
WaDelma: I would probably transform these to
```rust
let mut v = self.reads
    .iter()
    .flat_map(|stages| stages)
    .flat_map(|groups| groups)
    .cloned()
    .collect::<Vec<_>>();
```
or even:
```
let mut v = self.reads
    .iter()
    .flatten()
    .flatten()
    .cloned()
    .collect::<Vec<_>>();
```

src/dispatch/builder.rs
```diff
@@ -93,10 +93,10 @@ use crate::{
 pub struct DispatcherBuilder<'a, 'b> {
     current_id: usize,
     map: HashMap<String, SystemId>,
-    stages_builder: StagesBuilder<'a>,
+    pub(crate) stages_builder: StagesBuilder<'a>,
     thread_local: ThreadLocal<'b>,
     #[cfg(feature = "parallel")]
-    thread_pool: Option<::std::sync::Arc<::rayon::ThreadPool>>,
+    thread_pool: ::std::sync::Arc<::std::sync::RwLock<ThreadPoolWrapper>>,
```
WaDelma: There is an `Arc` inside `ThreadPoolWrapper`. I assume that the outer `Arc` is still needed, right?

examples/batch_dispatching.rs
AndreaCatania: In theory yes, to show that is dispatched in sequence due to the dependency.
I'll change it since doesn't make sense and I already have a sequential dispatching in the tests.

src/dispatch/async_dispatcher.rs
```diff
@@ -62,17 +63,23 @@ where
...
+            .as_ref()
+            .unwrap()
+            .spawn(move || {
+                {
+                    let world: &World = inner.world.borrow();
+
+                    for stage in &mut inner.stages {
+                        stage.execute(world);
+                    }
                 }
```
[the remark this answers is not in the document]
AndreaCatania: It could be avoided, but this mean call the function `Self::create_thread_pool()` each time a dispatcher builder is created.

If there is nothing against it I can take rid of that Option.

src/dispatch/batch.rs
```diff
@@ -0,0 +1,418 @@
...
+        self.reads.clone()
+    }
+
+    fn writes(&self) -> Vec<ResourceId> {
+        self.writes.clone()
+    }
+}
+
+/// The `BatchUncheckedWorld` wrap an instance of the world.
+/// It's safe to be used only in the context of the `Batch`.
```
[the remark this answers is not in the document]
AndreaCatania: Yes you can use it where you want. But is important that you create the BatchAccessor properly.

What I can do is rename it and make more clear how to use it.

src/dispatch/batch.rs
AndreaCatania: The `BatchControllerSystem` may want to read / write directly some resources.

Since it always has the `BatchUncheckedWorld`, I need a way to specify which resource it will use directly.

So the `BatchSystemData` serve to specify the resource that the controller will use.

src/dispatch/batch.rs
AndreaCatania: It's necessary to implement it otherwise it can't be sent trough thread and doesn't compile.
It's safe because it dispatch its inner systems using a dispatcher which will take care of execute it in order.

Do you think that I should write this here?

src/dispatch/builder.rs
AndreaCatania: The outer Arc is necessary because the wrapper is shared across the dispatcher and its internal batches.
Doing so allow change the ThreadPool to the main dispatcher and all its batches.

The internal Arc is needed because the ThreadPool could come from someone that is not using the Wrapper, for example the Amethyst thread pool.

src/dispatch/dispatcher.rs
AndreaCatania: Check my previous comment.

src/dispatch/dispatcher.rs
AndreaCatania: I'll add a macro

examples/batch_dispatching.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn accessor<'c>(&'c self) -> AccessorCow<'a, 'c, Self> {
+        AccessorCow::Ref(&self.accessor)
+    }
+
+    fn setup(&mut self, world: &mut World) {
+        self.dispatcher.setup(world);
+    }
+}
+
+unsafe impl<'a, 'b> Send for CustomBatchControllerSystem<'a, 'b> {}
```
WaDelma: I actually missed this `impl` first time reading through. I would highlight it by adding comment explaining it.

src/dispatch/batch.rs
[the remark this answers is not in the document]
WaDelma: I am not sure what it takes to create `BatchAccessor` correctly and how I would use this with it, so I wouldn't even know how to correctly use this.

src/dispatch/batch.rs
WaDelma: Ah now I understand what you are saying... But I am still not sure why cannot you specify `BatchControllerSystem` resources in the `System` `impl`? There has to be a reason, but it's not apparent from the code or documentation.

src/dispatch/batch.rs
WaDelma: I would write explanation why it's safe to implement the trait. Is it the `BatchAccessor` or the `Dispatcher<'a, 'b>` un-Send? What makes them un-Send by themselves? And how does wrapping them to `DefaultBatchControllerSystem` fix this un-Sendness?

src/dispatch/async_dispatcher.rs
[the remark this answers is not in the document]
WaDelma: I think I am missing something here. You cannot just remove the scope?

src/dispatch/dispatcher.rs
WaDelma: I am not sure how the previous comment answers this. Amethyst or anyone else could still create `Option<::std::sync::Arc<::rayon::ThreadPool>>`. I don't see what value this new type adds as it doesn't restrict it's api (public field) and we're not implementing any traits for it (to get around orphan rules).
So instead of newtype it could just be:
```rust
type ThreadPoolWrapper = Option<::std::sync::Arc<::rayon::ThreadPool>>;
```

src/dispatch/batch.rs
WaDelma: Well currently it's not *that* dynamic as those fields are private and cannot thus be edited dynamically. I don't have any good names for it either.

src/dispatch/async_dispatcher.rs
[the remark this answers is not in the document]
AndreaCatania: oh Sorry I miss understood it. Removed

src/dispatch/batch.rs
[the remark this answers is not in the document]
AndreaCatania: Ok I've changed it with:

/// The `BatchUncheckedWorld` wrap an instance of the world.
/// You have to specify this as `SystemData` of the `BatchControllerSystem`

src/dispatch/batch.rs
AndreaCatania: Consider this code:
```
impl<'a> System<'a> for CustomBatchControllerSystem<'_, '_> {
    type SystemData = BatchUncheckedWorld<'a>;

    fn run(&mut self, data: Self::SystemData) {
        {
            // Drop before dispatch the inner dispatcher
            let _ts = data.0.fetch::<TomatoStore>();
        }
        println!("Batch execution");
        for _i in 0..3 {
            self.dispatcher.dispatch(data.0);
        }
    }
}
```
The main reason is that before to execute the inner dispatcher you want to drop all the fetched resources otherwise if even one of that resources will be fetched by the inner dispatcher it will panics.

The example already show how to use it, and write this implementation detail doesn't seems important for the "user" of the library.
What do you think?

src/dispatch/batch.rs
AndreaCatania: It's something that is tight with the `Batch` execution. Since it seems the most descriptive name I'm leaving it as is until we don't have a better idea.

src/dispatch/dispatcher.rs
AndreaCatania: The comment got sorted in the wrong way, so let me quote it here for clarity

> The outer `Arc` is necessary because the wrapper is shared across the dispatcher and its internal batches.
> Doing so allow change the `ThreadPool` to the main dispatcher and all its batches.
> 
> The internal `Arc` is needed because the `ThreadPool` could come from someone that is not using the Wrapper, for example the Amethyst thread pool.


Basically using a wrapper as I did we have this situation:

```
Dispatcher ━━━━━━━━━━━━━━━━┓
   ├ Batch ━━━━━━━━━━━━━━ Wrapper ━━━━━━━>ThreadPool 1
   ├ Batch ━━━━━━━━━━━━━━━┛┃
   └ Batch ━━━━━━━━━━━━━━━━┛
```
When I set a new `ThreadPool` in the `Dispatcher` I change the inner part of the wrapper and the result will be this:

```
Dispatcher ━━━━━━━━━━━━━━━━┓
   ├ Batch ━━━━━━━━━━━━━━ Wrapper ━━━━━━━>ThreadPool 2
   ├ Batch ━━━━━━━━━━━━━━━┛┃
   └ Batch ━━━━━━━━━━━━━━━━┛
```

------------------
If instead I use the type like you advice (which is what actually shred use) I'll be in this state:
```
Dispatcher ━━━━━━━━━━━━━━━━┓
   ├ Batch ━━━━━━━━━━━━━━ ThreadPool 1
   ├ Batch ━━━━━━━━━━━━━━━┛┃
   └ Batch ━━━━━━━━━━━━━━━━┛
```
If I set a new `ThreadPool` in the `Dispatcher` only the Dispatcher will have it:
```
Dispatcher ━━━━━━━━━━━━━━ ThreadPool 2
   ├ Batch ━━━━━━━━━━━━━━ ThreadPool 1
   ├ Batch ━━━━━━━━━━━━━━━┛┃
   └ Batch ━━━━━━━━━━━━━━━━┛
```

src/dispatch/dispatcher.rs
AndreaCatania: Rethinking about it I'm not convinced in use a macro or and helper function.
I don't see a real benefit in hide that part of code.

Also this is used only in two locations, and is difficult that will be used elsewhere.

So I prefer to leave it as is.

AndreaCatania:
@WaDelma I just did a commit to resolve the things that you pointed out

src/dispatch/builder.rs
```diff
@@ -272,11 +350,18 @@ impl<'a, 'b> DispatcherBuilder<'a, 'b> {
     pub fn build(self) -> Dispatcher<'a, 'b> {
         use crate::dispatch::dispatcher::new_dispatcher;
 
+        #[cfg(feature = "parallel")]
+        self.thread_pool
+            .write()
+            .unwrap()
+            .0
+            .get_or_insert(Self::create_thread_pool());
```
WaDelma: btw should these be `get_or_insert_with` so it doesn't create thread pool if there is already something in the `Option`?

src/dispatch/dispatcher.rs
WaDelma: Am I doing something wrong here:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c43559e03501eb998c1dd2e4ba82ee69

src/dispatch/batch.rs
WaDelma: Thank you for explaining this in detail! Now I think I understand the relation between these two. I think it needs more explanation: If I wanted to implement my own batch controller I think I would have confused with this which would have made it harder than necessary.

Maybe add a comment that says that if the child systems depend on the same things as the batch controller there would be panic without separating these resource requirements.

src/dispatch/builder.rs
```diff
@@ -177,6 +177,84 @@ impl<'a, 'b> DispatcherBuilder<'a, 'b> {
...
+        writes.extend(<T::BatchSystemData as SystemData>::reads());
+        writes.sort();
+        writes.dedup();
+
+        let accessor = BatchAccessor::new(reads, writes);
+        let dispatcher = dispatcher_builder.build();
+
+        let batch_system = T::create(accessor, dispatcher);
+
+        self.add(batch_system, name, dep);
```
WaDelma: Because you can construct `BatchAccessor`, call `create` from the trait and then manually call `self.add`, implementing `Send` batch controller is not safe.

src/dispatch/dispatcher.rs
AndreaCatania: You forgot to use `ThreadWrapperPool` check this https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8d0df726d3e72398a355453d95cc2ba4

src/dispatch/builder.rs
AndreaCatania: Correct, thanks!

src/dispatch/builder.rs
AndreaCatania: Make sense

src/dispatch/dispatcher.rs
WaDelma: And this does the same: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=349fcdcb498585cf0a4a4d8209b3c00b

src/dispatch/batch.rs
AndreaCatania: I've added this to the `BatchSystemData` what do you think?

    /// Here you must set all the `Resource` types that you want to use directly
    /// inside the `BatchControllerSystem`.
    /// 
    /// These have to be specified here, instead of `SystemData` (as
    /// a normal `System` does) because there is a direct dependecy with the sub `System`s
    /// `Resource`.
    /// 
    /// Since the sub `System`s will use the same `Resource' of the `BatchControllerSystem`
    /// is necessary to drop the references to the fetched `Resource` before to
    /// dispatch the inner `Dispatcher`.
    /// 
    /// Now is easy to understand that specify the `BatchControllerSystem` `Resource`
    /// in the `SystemData` doesn't allow to drop the reference before the sub dispatching
    /// resulting in `Panic`.
    /// 
    /// So this mechanism allow you to specify the `Resource` that you will use in the `BatchControllerSystem`
    /// then is safe to fetch them directly from the world (As the example  "examples/batch_dispatching.rs" show).
    ///
    /// Note that it's not required to specify the sub systems resources here because they
    /// are handled automatically.
    type BatchSystemData: SystemData<'a>;

AndreaCatania:
@WaDelma another commit done to fix the last pointed things.

src/dispatch/dispatcher.rs
AndreaCatania: Correct, let me change it!

AndreaCatania: Done!

src/dispatch/batch.rs
```diff
@@ -0,0 +1,453 @@
...
+    /// a normal `System` does) because there is a direct dependecy with the sub
+    /// `System`s `Resource`.
+    ///
+    /// Since the sub `System`s will use the same `Resource' of the
+    /// `BatchControllerSystem` is necessary to drop the references to the
+    /// fetched `Resource` before to dispatch the inner `Dispatcher`.
+    ///
+    /// Now is easy to understand that specify the `BatchControllerSystem`
+    /// `Resource` in the `SystemData` doesn't allow to drop the reference
+    /// before the sub dispatching resulting in `Panic`.
```
WaDelma: I would write something like:
```markdown
Because `System::SystemData` is fetched for the whole duration of `System::run` dropping the resources isn't possible and dispatching inner `System`s will result to panic.
```
Except you can actually use `drop` function to drop the resources early... So do we actually need this separation? Or am I missing something here again.

Also you have been using the `BatchControllerSystem` term which doesn't exist in the code (which is why these suggestions), but if we still need the trait should we do `trait BatchDispatcherSystem: System` to make it a thing and force implementor to implement both which is what you need to any way do if you want to make custom batch dispatcher.

src/dispatch/batch.rs
```diff
@@ -0,0 +1,453 @@
...
+    ///
+    /// So this mechanism allow you to specify the `Resource` that you will use
+    /// in the `BatchControllerSystem` then is safe to fetch them directly
```
WaDelma: Is this "safe" here memory safe or panic safe? Because how safety is used in rust it should probably be specified.

src/dispatch/batch.rs
```diff
@@ -0,0 +1,453 @@
...
+    /// from the world (As the example  "examples/batch_dispatching.rs" show).
+    ///
+    /// Note that it's not required to specify the sub systems resources here
+    /// because they are handled automatically.
+    type BatchSystemData: SystemData<'a>;
+
+    /// Create the instance of the `BatchControllerSystem`
```
WaDelma:
```suggestion
    /// Creates an instance of the `BatchControllerSystem`
```

src/dispatch/batch.rs
```diff
@@ -0,0 +1,453 @@
...
+    ///
+    /// This function is unsafe because it depends from the `BatchAccessor` that
+    /// if created wrongly can Panics.
```
WaDelma: Panicking is not memory unsafe in general and so I am still not sure what is going on with this whole `Send` business. 
@torkleyy Is there reason for not implementing `Send` for `Dispatcher`?

examples/batch_dispatching.rs
```diff
@@ -0,0 +1,149 @@
...
+//!
+//! Specifically here we have three Systems
+//! - **SayHelloSystem**: Which is directly registered under the main
+//!   dispatcher.
+//! - **BuyTomatoSystem** and **BuyPotatoSystem** are registered in the batch.
+//!
+//! Notice that none of them is using any direct dependency.
+//! The SayHelloSystem is requesting the resources **TomatoStore** and
+//! **PotatoStore**, which are also requested by the other two systems inside
+//! the batch and by the batch controller itself.
```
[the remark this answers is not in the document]
WaDelma: I am not sure about this anymore.

WaDelma:
I also realise that the suggestion thing is not really suited for multiline suggestions and it makes it bit hard to read what I changed. Sorry about that. (Gitlab allows multiline suggestions)

examples/batch_dispatching.rs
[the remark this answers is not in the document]
AndreaCatania: This is saying that the `BatchControllerSystem` is using a resource that is also used by a sub `System` dispatched by the batch.

I think that controller is used correctly here

src/dispatch/batch.rs
```diff
@@ -0,0 +1,453 @@
...
+/// Is safe to implement `Send` and `Sync` because the `BatchAccessor` keep
+/// tracks of all used resources and thus the `System` can be safely executed in
+/// multi thread.
+pub trait BatchController<'a, 'b> {
+    /// Here you must set all the `Resource` types that you want to use directly
+    /// inside the `BatchControllerSystem`.
+    ///
+    /// These have to be specified here, instead of `SystemData` (as
+    /// a normal `System` does) because there is a direct dependecy with the sub
+    /// `System`s `Resource`.
```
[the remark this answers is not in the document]
AndreaCatania: I've updated it, now should make more sense

src/dispatch/batch.rs
AndreaCatania: You could need to fetch a resource each dispatching.

Using a `Drop` just before the dispatch, make the resource lost.
Then in the second iteration you have to fetch the resource from the world (which is not obvious).

The problem in this scenario is that the code is not uniform, and you have to make sure to fetch  only from the second iteration.

Instead I think that is more clear, uniform and flexible fetch it explicitly.

However I've updated the comment that should be more clear now.

src/dispatch/batch.rs
AndreaCatania: Is safe for both reason.
I mean that if you fetch an already fetched resource it panics, but also is wrong from memory prospective.

AndreaCatania:
Any news for this?

Cargo.toml
```diff
@@ -69,3 +69,7 @@ name = "seq_dispatch"
 
 [[example]]
 name = "thread_local"
+
+[[example]]
+name = "batch_dispatching"
+required-features = ["parallel"]
```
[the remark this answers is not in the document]
AndreaCatania: Done!

AndreaCatania:
Guys, thanks you a lot!

AndreaCatania:
I would like to update shred in Amethyst and so use this new feature.
@torkleyy can I help you to add a new version in crates.io?

jojolepro:
I'd just like to precise that I reviewed only half of the PR yesterday (lack of time), which is why I didn't add an "accept" review thingy. However, it as been reviewed by 2 people, so I would tend to think that I would not have found more than 1-2 nitpicks in the second half.

AndreaCatania:
@torkleyy, @jojolepro any idea about how to create a new version in crates.io?
otherwise I can't use it

jojolepro:
It depends when torkleyy actually wants to release. I'll ask him.

In the meantime, you can target the git repo using cargo using:
```
shred = { git = "https://github.com/slide-rs/shred" }
specs = { git = "https://github.com/your_specs_fork" }
```
For that, you'll need to fork specs and update that Cargo.toml file too so that it targets this repo.

AndreaCatania:
@jojolepro Thanks for the advice! I wanted to avoid it but that's ok, I can continue to work on phythyst now.

PR #149, opened 2019-07-30
[WIP] Expose `SystemExecSend` type alias.

azriel91:
This will be used by Amethyst to defer `System` instantiation.


azriel91:
actually, it doesn't satisfy `rustc`'s constraints yet, let me work out the end-to-end solution first

azriel91:
This is not necessary

PR #150, opened 2019-08-02
Change wording in BatchController and impose Send requirements

WaDelma:
Because panicking is general not unsafe, changes the doc comment of `BatchController::create` to better match the doc comment of the trait regarding safety.

As the implementers of the trait are supposed to be `Send` I added the requirement explicitly. 

I am still not sure why `Dispatcher` isn't `Send`, so I am not sure about the safety considerations.

### DIFF
--- a/src/dispatch/batch.rs
+++ b/src/dispatch/batch.rs
@@ -54,7 +54,7 @@ impl<'a> DynamicSystemData<'a> for BatchUncheckedWorld<'a> {
 /// The safety of implementing `Send` is ensured by `BatchAccessor` which keeps
 /// tracks of all used resources and thus the `System` can be safely executed in
 /// with multiple threads.
-pub trait BatchController<'a, 'b>: Send {
+pub trait BatchController<'a, 'b> {
     /// This associated type has to contain all resources batch controller uses directly.
     ///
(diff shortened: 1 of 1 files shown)


WaDelma:
Now that I think about it adding `Send` bound is probably not good idea.

src/dispatch/batch.rs
```diff
@@ -68,20 +68,23 @@ pub trait BatchController<'a, 'b> {
     /// `Resource` in the `SystemData` doesn't allow to drop the reference
     /// before the sub dispatching; resulting in `Panic`.
     ///
-    /// So this mechanism allow you to fetch safelly the specified `Resource`
+    /// So this mechanism allows you to fetch safely the specified `Resource`
     /// in the `BatchController`.
-    /// The example "examples/batch_dispatching.rs" show how to.
+    /// The example "examples/batch_dispatching.rs" show how to us it.
```
azriel91: Typo: us -> use

WaDelma:
The `Send` bound being bad idea is that it restricts what use can do: If they are using `BatchController` in single threaded environment `Send` bound just adds extra burden and it doesn't really add any extra value for multi threaded environment.

PR #163, opened 2019-09-11
Fix building without the "parallel" feature

mystal:
Seems like this was broken between `0.9.2` and `0.9.3`.

I'm not at all familiar with this code base, so please let me know if there is a better way to do this. That said, this compiles and unblocks my project locally.

### DIFF
--- a/src/dispatch/builder.rs
+++ b/src/dispatch/builder.rs
@@ -4,12 +4,14 @@ use hashbrown::HashMap;
 
 use crate::{
     dispatch::{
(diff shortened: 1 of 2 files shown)


TatriX:
We desperately need this to work. What can I do to help?

azriel91:
Heya, I'm going to merge this and release a `0.9.4`, but shortly after (within a day) am going to release a `0.10.0` for the `"nightly"` feature removal (version bump to not break crates that reference the `"nightly"` feature).

bors r+

PR #166, opened 2019-10-04
cargo build --no-default-features fails

bbodi:
- Fix the issue that it is not possible to compile shred without parallel feature, e.g. `cargo build --no-default-features` fails.


bbodi:
Sorry, duplicate of https://github.com/slide-rs/shred/pull/163

PR #170, opened 2019-11-04
Create ci-script.sh

torkleyy:


torkleyy:
Alright, this still needs a chmod 777, which I cannot do right now. After that, CI should actually fail correctly...

PR #183, opened 2019-12-31
Improvement/182/better fetch panic message

### LINKED ISSUE #182: Better fetch panic message
Currently when fetching a resource which has not been inserted, the panic message may be difficult for consumers to understand:

```
Tried to fetch a resource of type "amethyst::ecs::storage::MaskedStorage<project::CustomComponent>", but the resource does not exist.
Try adding the resource by inserting it manually or using the `setup` method.
```

The following should be more helpful:

```
Tried to fetch resource of type `MaskedStorage<CustomComponent>`[^1] from the `World`, but the resource does not exist.

You may ensure the resource exists through one of the following methods:

* Inserting it when the world is created: `world.insert(..)`.
* If the resource implements `Default`, include it in a system's `SystemData`, and ensure the system is registered in the dispatcher.
* If the resource does not implement `Default`, insert in the world during `System::setup`.

[^1]: Full type name: `amethyst::ecs::storage::MaskedStorage<project::CustomComponent>`
```

azriel91:
Closes #182.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,11 @@
 * Updated `smallvec` from `0.6` to `1.1`. ([#180])
 * `shred-derive`: Updated `syn`, `quote`, `proc-macro2` to `1.0`. ([#176])
 * ***Minimum Supported Rust Version updated to `1.38.0`***. ([#176])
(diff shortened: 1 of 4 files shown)


azriel91:
Thanks!

PR #186, opened 2020-01-24
Remove Send+Sync resource requirement if parallel disabled

chinedufn:
https://github.com/amethyst/specs/issues/673


chinedufn:
Don't mind at all - thanks a lot!

PR #209, opened 2021-06-07
Update dependency versions.

azriel91:


azriel91:
Rust 1.38.0 check fails because `arrayvec` needs minimum Rust 1.51.0 for const generics.

Won't get this merged then -- not sure if we want to raise the MSRV.

PR #215, opened 2022-05-16
Add getters to DispatcherBuilder

SebastienGllmt:
For our system graph, it's possible that in the end no system actually needs to be executed at the  given moment. To avoid the overhead of calling ` dispatcher.dispatch(&world)` on an empty graph, I wanted to add an `is_empty` function that I can call before executing.

Similar, some systems may not need to run at some given time.  There are times in my code where a child task needs to run even if its dependency did not need to execute, and there are times where it doesn't need to run. In order to implement this logic, I added a `has_system` function so that I can check whether or not a dependency was skipped

### DIFF
--- a/src/dispatch/builder.rs
+++ b/src/dispatch/builder.rs
@@ -112,6 +112,23 @@ impl<'a, 'b> DispatcherBuilder<'a, 'b> {
         Default::default()
     }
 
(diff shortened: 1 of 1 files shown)


torkleyy (approved):
:shipit:

PR #218, opened 2022-07-10
Eliminate safety issue in Resource::downcast*

### LINKED ISSUE #217: Update mopa to avoid safety issue (no longer maintained)
Dependabots recently opened an advisory on one of my projects https://github.com/Cypher1/tako/ because I'm using specs, which depends on shred, which depends on a version of mopa, whose latest version contains an unsound pattern which can cause segfaults.

Issue here:
https://github.com/chris-morgan/mopa/issues/13

Currently the thread's contributors state that mopa is not maintained, but one has set up mopa-maintained, as an alternative where they have added a fix.

Please investigate and update your deps so we can all avoid potential for segfaults :)

Cheers!

torkleyy:
Fixes #217

- The previous implementation uses `mopa`, which has a known safety
issue: https://github.com/chris-morgan/mopa/issues/13
- This commit essentially inlines the code generated by
the `mopafy!` macro provided by `mopa`, thus removing the
dependency
- Since `mopa::Any` was part of the public interface of the `Resource`
trait, this is technically a breaking change

Thank you to @Cypher1 for reporting this issue

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,6 @@ travis-ci = { repository = "amethyst/shred" }
 [dependencies]
 arrayvec = "0.7.2"
 hashbrown = "0.12"
-mopa = "0.2.2"
 rayon = { version = "1.5.0", optional = true }
 shred-derive = { path = "shred-derive", version = "0.6.3", optional = true }
 smallvec = "1.6.1"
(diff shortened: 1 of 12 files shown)


Cypher1:
Thanks! Lgtm

PR #220, opened 2022-07-13
Use ahash instead of SipHash

torkleyy:
Partially reverts 496564035390f085f677ed4e37852b1a5b4854cd (we use ahash again, but without using `hashbrown` which is just a copy of `std::collections::HashMap` with `ahash` as the default).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
       - master
   pull_request:
     branches:
-      - '**'
+      - "**"
 
 name: CI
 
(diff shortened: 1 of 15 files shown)


torkleyy:
cc @xMAC94x 

This gets us back to the speed before I removed `hashbrown` in the last PR, but without the unnecessary dependency on `hashbrown`.

xMAC94x:
mhhh okay, do we have any benchmarks or how did we compare the performance regarding to hasbrown ?

torkleyy:
The benchmarks are linked in your comment:

> An attempt to switch from fxhash to ahash resulted in [slowdowns of 1-4%](https://github.com/rust-lang/rust/issues/69153#issuecomment-589504301).

I think benchmarks in `shred` for this would be great as well, but for now, I just want to revert the accidental performance regression. Once we have such benchmarks we can also consider other hashers if they show beneficial.

PR #222, opened 2022-12-13
fix #221 - have DispatcherBuilder::add_batch() use the correct access

### LINKED ISSUE #221: Potential bug with DispatcherBuilder::add_batch()
I don't know enough about the inner workings of shred to know if this is actually a bug, but it looks like one.

DispatcherBuilder::add_batch() has what looks like a [copy/paste error](https://github.com/amethyst/shred/blob/master/src/dispatch/builder.rs#L278):
```
let mut reads = dispatcher_builder.stages_builder.fetch_all_reads();
reads.extend(<T::BatchSystemData as SystemData>::reads());
reads.sort();
reads.dedup();

let mut writes = dispatcher_builder.stages_builder.fetch_all_writes();
writes.extend(<T::BatchSystemData as SystemData>::reads());                 <<<<< here
writes.sort();
writes.dedup();

let accessor = BatchAccessor::new(reads, writes);
```

When determining the write accesses for the batch being added, it calls `reads()` instead of `writes()`. It seems like this could pretty negatively affect the ability to parallelize things around the batch.

I would make a PR, but I wasn't sure if there was actually a reason for doing things this way or not.

shimaowo:


shimaowo:
In case this actually is a bug.

torkleyy (approved):

Imberflur (approved):
Reviewing the git history of this line, I don't see any reason there that this isn't a bug.

PR #223, opened 2023-02-05
TrustCell fixes

Imberflur:
Fixing some unsoundness, see commit messages.

I'm hoping to eventually [replace TrustCell with atomic_refcell::AtomicRefCell](https://github.com/amethyst/shred/pull/224), however that should probably wait on https://github.com/bholley/atomic_refcell/pull/18

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     timeout-minutes: 10
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
@@ -58,3 +58,17 @@ jobs:
         with:
           command: test
           args: --verbose
+
+  miri:
+    name: "Miri"
+    runs-on: ubuntu-latest
+    continue-on-error: true # Needed until all Miri errors are fixed
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Miri
+        run: |
+          rustup toolchain install nightly --component miri
+          rustup override set nightly
+          cargo miri setup
+      - name: Test with Miri
+        run: cargo miri test
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -232,9 +232,9 @@ fn bench_fetching(b: &mut Bencher) {
 
     b.iter(|| {
         for _ in 0..100 {
-            world.fetch::<DeltaTime>();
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -58,3 +58,17 @@ jobs:
         with:
           command: test
           args: --verbose
+
+  miri:
+    name: "Miri"
+    runs-on: ubuntu-latest
+    continue-on-error: true # Needed until all Miri errors are fixed
```
Imberflur, line 65: Added Miri to the CI, however it still runs into some errors in the `MetaTable` code that is handling vtable pointers. So I set `continue-on-error: true` for now (hopefully this does what I think it does...)

benches/bench.rs
```diff
@@ -232,9 +232,9 @@ fn bench_fetching(b: &mut Bencher) {
 
     b.iter(|| {
         for _ in 0..100 {
-            world.fetch::<DeltaTime>();
-            world.fetch::<VecStorage<Pos>>();
-            world.fetch::<VecStorage<Spring>>();
+            black_box(world.fetch::<DeltaTime>());
+            black_box(world.fetch::<VecStorage<Pos>>());
+            black_box(world.fetch::<VecStorage<Spring>>());
```
Imberflur, line 237: I added `black_box` just to be more sure that the compile doesn't optimize these out (didn't really see much difference in the benchmark though)

src/cell.rs
```diff
@@ -613,4 +636,18 @@ mod tests {
...
+    fn drop_and_borrow_in_fn_call() {
+        fn drop_and_borrow(cell: &TrustCell<u8>, borrow: Ref<'_, u8>) {
+            drop(borrow);
+            *cell.borrow_mut() = 7u8;
+        }
+
+        let a = TrustCell::new(4u8);
+        let borrow = a.borrow();
+        drop_and_borrow(&a, borrow);
+    }
```
Imberflur, line 652: I added this to reproduce the scenario in the issue linked in the comment, however Miri was already reporting UB due to retagging of the reference by the `mem::forget` call in `RefMut::map`.

Imberflur (commented):
Here are some notes to help reviewers

.github/workflows/ci.yml
Imberflur: Well it seems like there is no way to not show the red `X` in the PR list: https://github.com/orgs/community/discussions/15452

PR #224, opened 2023-02-05
Replace TrustCell implementation with atomic_refcell::AtomicRefCell

Imberflur:
This PR replaces `TrustCell` with [`atomic_refcell::AtomicCell` ](https://docs.rs/atomic_refcell/latest/atomic_refcell/struct.AtomicRefCell.html)which provides near identical functionality.

Advantages of this change:
* `atomic_refcell` uses an alternative strategy to handle overflows of the read counter which allows for a single atomic operation when acquiring a read guard.
*  Removing the `TrustCell` implementation reduces the amount of `unsafe` code in this crate.

~~Based on top of commits from: https://github.com/amethyst/shred/pull/223~~
~~Waiting on: https://github.com/bholley/atomic_refcell/pull/18~~


src/world/mod.rs
```diff
@@ -542,15 +540,23 @@ impl World {
...
 
     /// Internal function for fetching resources, should only be used if you
     /// know what you're doing.
-    pub fn try_fetch_internal(&self, id: ResourceId) -> Option<&TrustCell<Box<dyn Resource>>> {
+    ///
+    /// # Safety
+    ///
+    /// If this is used to replace the `Box<dyn Resource>` with a different one, the new one must
+    /// have a `TypeId` that matches the one in the `ResourceId` provided here.
+    pub unsafe fn try_fetch_internal(
```
Imberflur, line 556: Oh almost forgot!

The one other thing I did in these changes is to mark this method as `unsafe` since it would otherwise allow safely swapping out the `Box<dyn Resource>` stored at a particular `ResourceId` with any other arbitrary `Box<dyn Resource>` that doesn't necessarily correspond to the `ResourceId`.

torkleyy (approved):
Looks good to me! Good catch on the unsafe function.

PR #225, opened 2023-02-11
Add missing Send/Sync impls

Imberflur:
Forgot these in https://github.com/amethyst/shred/pull/223!


src/cell.rs
```diff
@@ -47,6 +47,10 @@ pub struct Ref<'a, T: ?Sized + 'a> {
     value: NonNull<T>,
 }
 
+// SAFETY: `Ref<'_, T> acts as a reference.
+unsafe impl<'a, T: ?Sized + 'a> Sync for Ref<'a, T> where T: Sync {}
```
zesterer, line 51: I think this is correct, but can we express the bounds as `where for<'a> &'a T: Sync` and `where for<'a> &'a T: Send` instead so we're matching the impls with those of actual references?

Imberflur: Sure!

Imberflur: Amended commit

Imberflur: Actually I would like to double check if this works when `T` is not `'static`

Imberflur: seems fine

zesterer:
Miri failure? Hopefully nothing important

Imberflur:
> Miri failure? Hopefully nothing important

This is to be [expected for now](https://github.com/amethyst/shred/pull/223#discussion_r1096625095). `MetaTable` does some very questionable stuff with vtable pointers. I haven't ran into tests (in `specs`) that fail Miri due to this yet so I'm not focusing on it (but I would like to have reminder of Miri failing). Unfortunately, github CI lacks ability to have a CI check that is allowed to fail without showing red X <https://github.com/orgs/community/discussions/15452>. IIRC I think `MetaTable` is used for potential scripting support.

PR #232, opened 2023-11-20
Fix paths in SystemData macro

IsseW:


Imberflur:
Iirc this is commonly used via a re-export from `specs` so many users won't have the `shred` crate in scope. I think it is worth giving this case some consideration.

IsseW:
> Iirc this is commonly used via a re-export from `specs` so many users won't have the `shred` crate in scope. I think it is worth giving this case some consideration.

Through specs you have `specs::shred` to import in this case. Which I think will be a better thing to have to import than all of `World`, `SystemData` and `ResourceId`. Which could collide with user code.

So maybe if there aren't leading any `::` this could work? I don't think there's a way to solve not having to import anything from specs without creating a seperate proc macro in specs.

IsseW:
Removed leading `::`.

src/system.rs
```diff
@@ -284,6 +284,11 @@ pub trait SystemData<'a> {
     ///
     /// Please note that returning wrong dependencies can lead to a panic.
     fn writes() -> Vec<ResourceId>;
+
+    /// Returns names of retrieved resources.
+    fn resource_names() -> Vec<&'static str> {
+        vec![std::any::type_name::<Self>()]
+    }
```
Imberflur, line 291: I just noticed that this PR is adding a new method to the SystemData trait. What is the purpose here? Was it intended to be included in this PR?

IsseW: Oh huh, I hadn't even noticed that. Must be from some testing a long time ago

Imberflur (approved):

PR #234, opened 2024-01-02
Add a dispatcher that can be sent between threads

Imberflur:


IsseW:
`Dispatcher` seems to now be an extension over `SendDispatcher` if IIUC.

And I think there would be less code duplication if `Dispatcher` contained a `SendDispatcher` instead of the two fields they share.

Imberflur:
> `Dispatcher` seems to now be an extension over `SendDispatcher` if IIUC.
> 
> And I think there would be less code duplication if `Dispatcher` contained a `SendDispatcher` instead of the two fields they share.

implemented this!

IsseW (approved):

PR #236, opened 2024-05-09
Fix compilation on latest nightly

crabm4n:
`core::ptr::invalid_mut` from the [`strict_provenance`](https://github.com/rust-lang/rust/issues/95228) nightly feature seems to have been silently renamed to [`core::ptr::without_provenance_mut`](https://doc.rust-lang.org/nightly/std/ptr/fn.without_provenance_mut.html) breaking `vek` on the latest nightly toolchain.


Imberflur:
Thanks!

Imberflur (approved):

PR #237, opened 2024-05-13
Update aHash / tynm

xMAC94x:
update aHash because of their license dependency and update tynm because their minimum version no longer builds in future rust compilers


Imberflur (approved):
looks good!

PR #239, opened 2025-02-14
Print type name when panicking due to violating aliasing laws

richard-uk1:
When you have a lot of storages in a big project, and especially when backtraces are not available (e.g. web) it can be difficult to know which storage is breaking the borrow rules (mut + immut borrow or 2+ mut borrows). This patch adds the type name to the panic message.

### DIFF
--- a/src/world/mod.rs
+++ b/src/world/mod.rs
@@ -426,9 +426,15 @@ impl World {
         T: Resource,
     {
         let res_id = ResourceId::new::<T>();
+        let resource = self.resources.get(&res_id)?;
 
-        self.resources.get(&res_id).map(|r| Fetch {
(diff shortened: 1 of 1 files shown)


Imberflur:
Hi, was this closed from inactivity? I just fixed the unrelated issues causing the CI to fail.

richard-uk1:
No, I'm no longer working at the company that this work was for. But I can re-open if you want to take the PR?

Imberflur:
> But I can re-open if you want to take the PR?

Looks like a nice improvement to me and didn't have any performance impact when I tested it so I think merging it would be nice.

richard-uk1:
CI is green

Imberflur:
Thanks!

PR #240, opened 2025-02-21
Fix CI errors

Imberflur:


torkleyy (commented):
I think the minimum rust version should be bumped in the README :D

